PR #1, opened 2015-08-26
Create flay-based engine

jacobi007:


.codeclimate.yml
```diff
@@ -0,0 +1,6 @@
+engines:
```
GordonDiggs: This should probably have rubocop too

Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM alpine:edge
```
GordonDiggs: This is missing some items from the [spec](https://github.com/codeclimate/spec/blob/master/SPEC.md#packaging)

lib/cc/engine/flay.rb
```diff
@@ -0,0 +1,57 @@
+require 'flay'
+require 'json'
+
+module CC
+  module Engine
+    class Flay
+      def initialize(directory: ,engine_config: ,io:)
```
GordonDiggs: :lipstick: spaces between the arguments: `directory:, engine_config:, io:`

lib/cc/engine/flay.rb
```diff
@@ -0,0 +1,57 @@
...
+        @directory = directory
+        @engine_config = engine_config
+        @io = io
+      end
+
+      def run
+        files = ::Flay.filter_files(::Flay.expand_dirs_to_files(directory))
```
GordonDiggs: Does this handle exclude_paths?

lib/cc/engine/flay.rb
```diff
@@ -0,0 +1,57 @@
...
+        flay.process(*files)
+
+        flay.report(StringIO.new).each do |issue|
+          puts "#{new_violation(issue).to_json}\n"
```
GordonDiggs: These should be separated with `\0`, instead of `\n`. JSON can contain newlines so we use the null byte to be safer.

Dockerfile
jacobi007: Could you be a bit more specific? Also, should I set codeclimate as a maintainer?

GordonDiggs: Yep that maintainer sounds good. It looks good now save for setting the WORKDIR as `/code`

PR #2, opened 2015-09-08
Add duplication engine analyzing ruby code

jacobi007:


lib/cc/engine/analyzers/ruby/main.rb
```diff
@@ -0,0 +1,54 @@
+require 'flay'
+require 'json'
+require 'cc/engine/analyzers/helpers/main'
+require 'pry'
```
pbrisbin: accidental?

lib/cc/engine/analyzers/ruby/main.rb
```diff
@@ -0,0 +1,54 @@
...
+
+          attr_reader :directory, :engine_config, :io
+
+          def filter_files(files)
+            return files if engine_config["exclude_paths"].nil?
+            files - excluded_files
+          end
+
+          def mass_threshold
+            mass = engine_config.fetch('config', {}).fetch('ruby', {}).fetch('mass_threshold', {})
```
pbrisbin: This seems to default to `{}` them set as `10` if `empty?`. Would it make more sense to use `fetch("mass_threshold", 10)`?

mrb:
@codeclimate/review

Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM alpine:edge
```
wfleming: Should we use our internal tagged alpine image? Or does alpine themselves provide any tags besides "edge"? We've been bitten a couple times now by base images changing in ways that broke our images, so I feel like we should avoid relying on third-party "latest"/"edge" images.

Gemfile
```diff
@@ -0,0 +1,14 @@
+source 'https://rubygems.org'
+
+gem 'flay'
+gem 'pry'
+gem 'posix-spawn'
+gem 'sexp_processor'
+gem 'json'
+gem 'mocha'
```
wfleming: nit, but this looks like it could go in the `:test` group?

lib/cc/engine/analyzers/helpers/main.rb
```diff
@@ -0,0 +1,73 @@
...
+              "end": location.line
+            }
+          }
+        end
+
+        def format_locations(other)
+          other.map { |location| format_location(location) }
+        end
+
+        def options
```
wfleming: nit, but since you use this module as an include, I could see a day when somebody includes it in a runner-type class that implements it's own `options` method, and confusion ensues. `flay_options` might be clearer?

PR #3, opened 2015-09-24
Add php support

jacobi007:


lib/cc/engine/analyzers/php/parser.rb
```diff
@@ -0,0 +1,70 @@
...
+          def run(input, timeout = DEFAULT_TIMEOUT)
+            Timeout.timeout(timeout) do
+              child = ::POSIX::Spawn::Child.new(command, input: input, timeout: timeout)
+              if child.status.success?
+                output = block_given? ? yield(child.out) : child.out
+                delegate.on_success(output)
+              end
+
+            end
+          rescue *EXCEPTIONS
```
pbrisbin: Is the intention really to quietly ignore these?

jacobi007: I've reused the code from the old platform and not removing the `rescue` seemed to be a good idea. In the past, the rescue block was used to store values like timestamp on Statsd object, which at the moment is not included in the project.

pbrisbin: Gotcha. The situation is slightly different in engines land:

The engine itself shouldn't be concerned with exception handling (unless there are expected/ignorable exceptions) or timeouts. The application that's running the engine will handle all of that.

In short, I think you can remove this `rescue` and any timeout-related logic.

BlakeWilliams:
@pbrisbin Is this good to merge or should I open a new PR with the `rescue`s removed?

jacobi007:
@pbrisbin @BlakeWilliams I've removed the rescue/timeout blocks in the PHP module. Please note that the same logic resides in Javascript module as well but I think it should be removed in a separated PR.

BlakeWilliams:
@pbrisbin Merging this unless you have any objections.

@jacobi007 Want to squash or is this good to go?

pbrisbin:
@BlakeWilliams fine with me. To me, new engine PRs arent' super valuable to review closely because it's so much green. As long as they go through appropriate QA, I'd rather merge it and focus any code review on the bugfix PRs that come out after.

BlakeWilliams:
Sounds good, I'll get this merged in soon then.

PR #4, opened 2015-10-01
Add bin/setup to install esprima

BlakeWilliams:
The Node duplicaiton analyzer depends on `esprima` but isn't explicit
about this being a requirement or provide a setup script to get the
environment properly setup.

This adds a basic bin/setup script to install esprima and bundle gems.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+node_modules
--- a/bin/setup
+++ b/bin/setup
@@ -1,0 +1,6 @@
+#!/usr/bin/env sh
+
+set -f
+
+npm install esprima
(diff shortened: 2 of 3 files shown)


wfleming:
I notice we do an `npm install -g` in the `Dockerfile`. Could we centralize this dependency by putting it in a `package.json` & having both `bin/setup` & the `Dockerfile` just run `npm install`?

BlakeWilliams:
I think a `package.json` makes sense. I'll add that and have docker use `bin/setup`.

BlakeWilliams:
This is no longer relevant.

PR #5, opened 2015-10-05
Add python duplication support

BlakeWilliams:
This PR adds Python duplication support and fixes a few existing bugs to get this application into a good place.

* Adds ability to transform Python -> AST -> JSON -> S-Expression for flay.
* Runs `Flay#report` only once now to fix exponential issue bug.
* Only grabs the first reported location from flay and stores the rest as `other_locations`.
* Now returns `end` line instead of the current line for both beginning and end.
* Uses `_type` in the Python script to return JSON-ified AST since `type` is used by the AST.

A lot of this code is copied over from The JavaScript analyzer and the Python analyzer (from another repo). There's a good chance to refactor a lot of this but this PR is already doing a lot so I'd prefer to do that later.


BlakeWilliams:
@codeclimate/review I think this is ready for a review.

pbrisbin:
Looks good enough to me to move onto to trying it out.

lib/cc/engine/analyzers/helpers/main.rb
```diff
@@ -6,19 +6,30 @@ module Analyzers
...
+
         def flay
           @flay ||= ::Flay.new(flay_options)
         end
 
-        def new_violation(issue, location, other)
+        def report
+          flay.report(StringIO.new).each do |issue|
+            location = issue.locations.first
+            io.puts "#{new_violation(issue, location).to_json}\0"
```
joshuaclayton: where does `io` come from - something `each(&block)` provides on the report?

lib/cc/engine/analyzers/helpers/main.rb
```diff
@@ -35,18 +46,42 @@ def find_other_locations(all_locations, current)
...
-        def format_location(location)
+        def format_location(issue, location)
+          current_sexp = flay.hashes[issue.structural_hash].detect do |sexp|
+            sexp.line == location.line
+          end
+
+          format_sexp(current_sexp)
+        end
+
+        def format_locations(issue, location)
```
mxie: Could we do something where we're doing the split between `location` and `other_locations` in a single place and then have another place loop through and format? This method and the previous one looks too similar...

Here's something crazy:

```ruby
def new_violation(issue, location)
  first_sexp, other_sexps = formatted_sexps_by(location)

  {
    #...,
    "location": first_sexp,
    "other_locations": other_sexps,
  }
end

def formatted_sexps_by(location)
  flay.hashes[issue.structural_hash].partition do |sexp|
    sexp.line == location.line
  end.flatten.map do |sexp|
    format_sexp(sexp)
  end
end
```

This is technically shorter, but probably sacrifices readability...

Anyway, not a blocker.

spec/cc/engine/analyzers/javascript/main_spec.rb
```diff
@@ -28,24 +29,14 @@ def run_engine(config = nil)
...
-      end
-
-      def third_issue
-        {"type":"issue","check_name":"Identical code","description":"Duplication found in expression_statement","categories":["Duplication"],"location":{"path":"#{@code}/foo.js","lines":{"begin":3,"end":3}},"remediation_points":450000,"other_locations":[{"path":"#{@code}/foo.js","lines":{"begin":1,"end":1}},{"path":"#{@code}/foo.js","lines":{"begin":2,"end":2}}], "content":{"body": read_up}}
-      end
-
-      def printed_issues
-        first_issue.to_json + "\0\n" + second_issue.to_json + "\0\n" + third_issue.to_json + "\0\n"
+      def printed_issue
+        issue = {"type":"issue","check_name":"Identical code","description":"Duplication found in expression_statement","categories":["Duplication"],"location":{"path":"foo.js","lines":{"begin":1,"end":1}},"remediation_points":450000, "other_locations":[{"path":"foo.js","lines":{"begin":2,"end":2}},{"path":"foo.js","lines":{"begin":3,"end":3}}], "content":{"body": read_up}}
```
mxie: Hm, could we format this or would that defeat the purpose?

lib/cc/engine/analyzers/javascript/parser.rb
```diff
@@ -23,6 +23,8 @@ def parse
           def on_success(output)
             parsed_json = JSON.parse(output, max_nesting: false)
             parsed_json.delete('sourceType')
+
+
```
mxie: Remove extra new lines.

lib/cc/engine/analyzers/python/main.rb
```diff
@@ -0,0 +1,51 @@
...
+          end
+
+          def run
+            files_to_analyze.each do |file|
+              start_flay(process_file(file))
+            end
+          end
+
+          def process_file(path)
+            Node.new(::CC::Engine::Analyzers::Python::Parser.new(File.binread(path), path).parse.syntax_tree, path).format
```
mxie: Could we split out the args onto new lines for `Parser`?

lib/cc/engine/analyzers/python/main.rb
```diff
@@ -0,0 +1,51 @@
...
+          end
+
+          def mass_threshold
+            engine_config.fetch('config', {}).fetch('python', {}).fetch('mass_threshold', 50)
```
mxie: 2 things:

1. I'd put each `.fetch` on it's own line for better readability.
2. Let's extract the threshold count as a constant.

spec/cc/engine/analyzers/python/main_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+    describe "#run" do
+      it "prints an issue" do
+
+        create_source_file("foo.py", <<-EOJS)
+print("Hello", "python")
+print("Hello", "python")
+print("Hello", "python")
+        EOJS
+
+
```
mxie: Extra new line

lib/cc/engine/analyzers/python/node.rb
```diff
@@ -0,0 +1,74 @@
...
+          end
+
+          private
+
+          def properties_to_sexps
+            valid_properties.map do |key, value|
+              if value.is_a?(Array)
+                create_sexp(key.to_sym, *Node.new(value, @file, @line).format)
+              elsif value.is_a?(Hash)
+                create_sexp(key.to_sym, *Node.new(value, @file, @line).format)
```
mxie: Since we're treating Arrays and Hashes the same way, could we combine the conditions?

lib/cc/engine/analyzers/python/node.rb
```diff
@@ -0,0 +1,74 @@
...
+          def create_sexp(*args)
+            Sexp.new(*args).tap do |sexp|
+              sexp.file = @file
+              sexp.line = @line
+            end
+          end
+
+          def valid_properties
+            @node.reject do |key, value|
+              value_empty = value == nil || value == {} || value == []
```
mxie: How about `value.nil? || value.empty?`?

BlakeWilliams: We need to keep empty strings.

lib/cc/engine/analyzers/helpers/main.rb
BlakeWilliams: Apparently all of these subclasses use the same constructor so I need to extract this.

spec/cc/engine/analyzers/javascript/main_spec.rb
BlakeWilliams: I definitely think this should be formatted.

lib/cc/engine/analyzers/python/parser.rb
```diff
@@ -0,0 +1,57 @@
...
+            runner = CommandLineRunner.new(python_env, self)
+            runner.run(code) do |ast|
+              json_ast = JSON.parse(ast)
+              @syntax_tree = json_ast
+            end
+
+            self
+          end
+
+          def python_env
```
mxie: Hmm...does this method name accurately describes this? Seems more like a command to me.

lib/cc/engine/analyzers/python/main.rb
BlakeWilliams: Unfortunately we can't. These being named was (seemingly) the cause of a memory issue.

.gitignore
```diff
@@ -0,0 +1 @@
+node_modules
```
mxie: Does this belong?

lib/cc/engine/analyzers/python/main.rb
BlakeWilliams: I think this is another common thing between each languages `main` class. I might save this for the refactor.

lib/cc/engine/analyzers/python/parser.rb
BlakeWilliams: Agreed, this is largely stolen from the JavaScript parser.

.gitignore
BlakeWilliams: Yeah, we rely on a node module to run the JavaScript analyzer.

lib/cc/engine/analyzers/python/node.rb
mxie: Bah!

joshuaclayton: `[nil, {}, []].include?(value)`?

lib/cc/engine/analyzers/helpers/main.rb
BlakeWilliams: This turns out to be an include unfortunately. I think I'm going to make flay/reporting its own class and use it in the analyzers instead of including it.

lib/cc/engine/analyzers/file_list.rb
```diff
@@ -0,0 +1,33 @@
...
+          @directory = directory
+          @engine_config = engine_config
+          @extension = extension
+        end
+
+        def files
+          files = Dir.glob("#{directory}/**/*.#{extension}").reject do |f|
+            File.directory?(f)
+          end
+          files - excluded_files
```
mxie: Can we do this as part of the `reject` condition?

lib/cc/engine/analyzers/file_list.rb
```diff
@@ -0,0 +1,33 @@
...
+        end
+
+        private
+
+        attr_reader :directory, :engine_config, :extension
+
+        def excluded_files
+          return [] if excluded_paths.nil?
```
mxie: How about having `excluded_paths` return `[]` as an alternative instead?

lib/cc/engine/analyzers/file_list.rb
BlakeWilliams: We can, but I think this makes the intention a bit more clear than included both conditionals. Thoughts?

mxie: That's fair. Could we at least either rename the method or the variable so that they're not the same though?

lib/cc/engine/analyzers/reporter.rb
```diff
@@ -0,0 +1,65 @@
+require 'cc/engine/analyzers/violation'
+
+module CC
+  module Engine
+    module Analyzers
+      class Reporter
+        def initialize(directory, language_strategy, io)
+          @directory = directory
+          @language_strategy = language_strategy
+          @mass_threshold = language_strategy.mass_threshold
```
mxie: I'd actually rather have this be its own private method. I generally don't like doing "special stuff" right in the constructor. WDYT?

BlakeWilliams: :+1:

lib/cc/engine/analyzers/violation.rb
```diff
@@ -0,0 +1,86 @@
...
+        def current_sexp
+          @location ||= hashes.first
+        end
+
+        def other_sexps
+          @other_locations ||= hashes.drop(1)
+        end
+
+        def name
+          issue.identical? ? 'Identical code' : 'Similar code'
```
mxie: Expand ternery?

BlakeWilliams: Good catch, I can't stand ternaries.

lib/cc/engine/duplication.rb
```diff
@@ -35,6 +38,7 @@ def any_enabled_languages?
 
       def languages
         engine_config['languages'] || ENABLED_BY_DEFAULT
+        ['python']
```
mxie: I'm confused...what's the use of the previous line if we're just returning this array?

BlakeWilliams: Oops, this is because yaml-config doesn't work yet.

BlakeWilliams: It's just a testing necessity.

mxie:
Had a few questions, but this looks very nice and clean!

BlakeWilliams:
Thanks!

lib/cc/engine/analyzers/python/main.rb
```diff
@@ -0,0 +1,48 @@
...
+module CC
+  module Engine
+    module Analyzers
+      module Python
+        class Main
+          DEFAULT_MASS_THRESHOLD = 50
+
+          def initialize(directory:, engine_config:)
+            @directory = directory
+            @engine_config = engine_config || {}
```
joshuaclayton: will `engine_config` ever be `nil` here?

lib/cc/engine/analyzers/python/main.rb
```diff
@@ -0,0 +1,48 @@
...
+          end
+
+          def run
+            files.map do |file|
+              process_file(file)
+            end
+          end
+
+          def mass_threshold
+            engine_config.fetch('config', {}).fetch('python', {}).fetch('mass_threshold', DEFAULT_MASS_THRESHOLD)
```
joshuaclayton: double quotes?

lib/cc/engine/analyzers/javascript/main.rb
```diff
@@ -8,52 +8,48 @@ module Engine
...
 
-          private
-
-          def js_parser
-            ::CC::Engine::Analyzers::Javascript::Parser
-          end
-
           def mass_threshold
-            engine_config.fetch('config', {}).fetch('javascript', {}).fetch('mass_threshold', 10)
+            engine_config.fetch('config', {}).fetch('javascript', {}).fetch('mass_threshold', DEFAULT_MASS_THRESHOLD)
```
joshuaclayton: double quotes?

lib/cc/engine/analyzers/python/main.rb
```diff
@@ -0,0 +1,48 @@
...
+          end
+
+          private
+
+          attr_reader :directory, :engine_config
+
+          def process_file(path)
+            Node.new(::CC::Engine::Analyzers::Python::Parser.new(File.binread(path), path).parse.syntax_tree, path).format
```
joshuaclayton: thoughts on breaking this apart into a few pieces to make it easier to grok?

lib/cc/engine/analyzers/python/node.rb
```diff
@@ -0,0 +1,72 @@
...
+          def set_default_line(default)
+            if has_line_number?
+              @line = @node["attributes"]["lineno"]
+            else
+              @line = default
+            end
+          end
+
+          def has_line_number?
+            @node.is_a?(Hash) && @node["attributes"] && @node["attributes"]["lineno"]
```
joshuaclayton: what else could this be besides a hash? should we check that it responds to `[]` instead? Thoughts on forcing a hash and using fetch with defaults?

lib/cc/engine/analyzers/python/main.rb
BlakeWilliams: Not sure. I think this could be nil if a repo doesn't have a config.

lib/cc/engine/analyzers/python/main.rb
BlakeWilliams: I think this has to be in this format because it fixed a memory issue.

lib/cc/engine/analyzers/python/node.rb
BlakeWilliams: Hash, Array, String, etc. It's from JSON.

Would `fetch` be nicer here since the default is nested two levels down?

lib/cc/engine/analyzers/file_list.rb
BlakeWilliams: Done!

lib/cc/engine/analyzers/reporter.rb
```diff
@@ -0,0 +1,67 @@
...
+
+module CC
+  module Engine
+    module Analyzers
+      class Reporter
+        def initialize(directory, language_strategy, io)
+          @directory = directory
+          @language_strategy = language_strategy
+          @io = io
+          @flay = Flay.new(flay_options)
```
joshuaclayton: should this be responsible for instantiating `Flay`? Thoughts about inject it (or an adapter to it) instead?

joshuaclayton: basically, it could be a `sexp_processor` that has a `process(sexp)` method on it, in case something later on takes the place of Flay

lib/cc/engine/analyzers/reporter.rb
```diff
@@ -0,0 +1,67 @@
...
+          report
+        end
+
+        def report
+          flay.report(StringIO.new).each do |issue|
+            io.puts "#{new_violation(issue).to_json}\0"
+          end
+        end
+
+        def process(*files)
```
joshuaclayton: can this get removed?

lib/cc/engine/analyzers/reporter.rb
```diff
@@ -0,0 +1,67 @@
...
+          sexps = language_strategy.run
+          sexps.each do |sexp|
+            process_sexp(sexp)
+          end
+
+          report
+        end
+
+        def report
+          flay.report(StringIO.new).each do |issue|
```
joshuaclayton: what's the significance of `StringIo.new` here? could this get moved out to a well-named method signifying what it's used for?

lib/cc/engine/analyzers/reporter.rb
joshuaclayton: looking at this now, it seems like we use `mass_threshold` from the strategy - possibly more coupled to flay than I'd thought originally, it seems

lib/cc/engine/analyzers/reporter.rb
```diff
@@ -0,0 +1,67 @@
...
+        end
+
+        def flay_options
+          {
+            diff: false,
+            mass: mass_threshold,
+            summary: false,
+            verbose: false,
+            number: true,
+            timeout: 10,
```
joshuaclayton: magic number? thoughts on extracting somewhere?

lib/cc/engine/analyzers/ruby/main.rb
```diff
@@ -1,48 +1,44 @@
...
-                io.puts "#{new_violation(issue, location, other_locations).to_json}\0"
-              end
+            files.map do |file|
+              process_file(file)
             end
           end
 
-          private
+          def mass_threshold
+            engine_config.fetch('config', {}).fetch('ruby', {}).fetch('mass_threshold', DEFAULT_MASS_THRESHOLD)
```
joshuaclayton: double quotes?

lib/cc/engine/analyzers/violation.rb
```diff
@@ -0,0 +1,90 @@
...
+        private
+
+        attr_reader :hashes, :directory
+
+        def current_sexp
+          @location ||= hashes.first
+        end
+
+        def other_sexps
+          @other_locations ||= hashes.drop(1)
```
joshuaclayton: what is `hashes.drop`?

lib/cc/engine/duplication.rb
```diff
@@ -21,13 +22,15 @@ def run
         languages.each do |lang|
           next unless language_supported?(lang)
```
joshuaclayton: instead of calling `next`, what about having `languages` return only supported languages?

joshuaclayton: thoughts on, instead of languages, storing a hash of language to class, or just an array of the supported analyzers, to avoid some of the class-oriented programming?

lib/cc/engine/analyzers/reporter.rb
BlakeWilliams: Yeah, it's a bit coupled but I think it makes sense.

I'll extract it into a private method which should make it a bit better.

lib/cc/engine/analyzers/reporter.rb
BlakeWilliams: Yep!

lib/cc/engine/analyzers/reporter.rb
BlakeWilliams: Just looked it up, `Flay#report` accepts an output stream so `flay.report(StringIO.new)` is capturing the output into that `StringIO` instance that we don't use instead of sending it to the default `$stdout`.

`Flay#report` returns an array of `Flay::Item`s which we then use to generate violations.

lib/cc/engine/analyzers/reporter.rb
BlakeWilliams: Good catch, I'll make it a constant.

lib/cc/engine/analyzers/violation.rb
BlakeWilliams: It's from `Flay#hashes` which gives you a hash of `hash_string -> [s-expression]`. We drop here because these are `other_locations` to report.

lib/cc/engine/duplication.rb
BlakeWilliams: @joshuaclayton Just pushed this change. One more look please?

PR #6, opened 2015-10-12
Use correct config path to languages

BlakeWilliams:
The existing code was using `engine_config["languages"]` which is
incorrect. This needs to point to `engine_config["config"]["languages"]`
to work properly.


BlakeWilliams:
cc @codeclimate/review and @mrb

mrb:
:+1:

PR #7, opened 2015-10-13
Make base point values language specific

BlakeWilliams:
This extracts the base points to be a responsibility of each analyzer to
provide since each language can have different needs based on the
generated s-expressions.

This also puts us in a good place if we decide to make points part of
the configuration.

### DIFF
--- a/lib/cc/engine/analyzers/python/main.rb
+++ b/lib/cc/engine/analyzers/python/main.rb
@@ -11,7 +11,10 @@ module CC
         class Main
           LANGUAGE = "python"
           DEFAULT_PATHS = ["**/*.py"]
-          DEFAULT_MASS_THRESHOLD = 50
+          DEFAULT_MASS_THRESHOLD = 40
+          BASE_POINTS = 1000
+          LANGUAGE = "python"
+          DEFAULT_PATHS = ["**/*.py"]
 
           def initialize(directory:, engine_config:)
             @directory = directory
@@ -26,6 +29,10 @@ module CC
 
           def mass_threshold
             engine_config.fetch("config", {}).fetch("python", {}).fetch("mass_threshold", DEFAULT_MASS_THRESHOLD)
+          end
+
+          def base_points
+            BASE_POINTS
           end
 
           private
--- a/lib/cc/engine/analyzers/python/node.rb
+++ b/lib/cc/engine/analyzers/python/node.rb
@@ -5,7 +5,7 @@ module CC
     module Analyzers
       module Python
         class Node < CC::Engine::Analyzers::Node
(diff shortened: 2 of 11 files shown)


lib/cc/engine/analyzers/python/node.rb
```diff
@@ -3,7 +3,7 @@ module Engine
     module Analyzers
       module Python
         class Node
-          SCRUB_PROPERTIES = ["_type", "attributes"].freeze
```
mxie: Is this a related change?

BlakeWilliams: Nope, this is me trying to trim down the resulting s-expressions since they have a heavy mass.

mxie: Can we make a separate commit for this instead then?

BlakeWilliams: Sure, I'll split it into it's own commit, but same PR.

mxie:
How do we use `BASE_POINTS` currently? And should we do something similar for `violation.rb`?

lib/cc/engine/analyzers/python/main.rb
```diff
@@ -9,7 +9,8 @@ module Engine
     module Analyzers
       module Python
         class Main
-          DEFAULT_MASS_THRESHOLD = 50
+          DEFAULT_MASS_THRESHOLD = 40
+          BASE_POINTS = 1000
```
GordonDiggs: Why are the base points different for python?

BlakeWilliams: The s-expressions generated from Python's AST is much more verbose than Ruby's which means the points have to be lower.

GordonDiggs: Cool - makes sense to me!

BlakeWilliams:
Re-review please? cc @mxie

lib/cc/engine/analyzers/python/node.rb
```diff
@@ -5,7 +5,7 @@ module Engine
     module Analyzers
       module Python
         class Node < CC::Engine::Analyzers::Node
-          SCRUB_PROPERTIES = ["_type", "attributes"].freeze
+          SCRUB_PROPERTIES = ["_type", "attributes", "ctx"].freeze
```
GordonDiggs: :lipstick: Can you make this a `%w[]` array?

BlakeWilliams: Yep! That'll look much nicer.

GordonDiggs:
One minor nit but otherwise LGTM

BlakeWilliams:
Thanks!

PR #8, opened 2015-10-15
Migrate from esprima to babel

BlakeWilliams:
This PR replaces the esprima ast with babel since babel seems to stay
more up to date and also gives us free JSX support.

This also let us extract `Node` from the Python analyzer into a common
base class. This works because all AST's *should* behave mostly the same
once turned into JSON with only minor differences here and there.


lib/cc/engine/analyzers/sexp.rb
```diff
@@ -0,0 +1,3 @@
+class Sexp
+  attr_accessor :end_line
```
BlakeWilliams: This is required to get the correct end of JavaScript expressions.

Without it, the following:

```
if (foo) {
  bar();
}
```

Would show up as:

```
if (foo) {
  bar();
```

In the application.

BlakeWilliams:
cc @codeclimate/review

GordonDiggs:
One thing I noticed yesterday is that we only support one file extension per language, but in order to get true JSX support, we need `js` and `jsx`. Do you have plans to address that?

lib/cc/engine/analyzers/javascript/parser.js
```diff
@@ -1,15 +1,51 @@
...
   source += chunk;
 });
 
 process.stdin.on('end', function() {
-  var result = esprima.parse(source, { loc: true });
-  process.stdout.write(JSON.stringify(result));
+  var ast = babel.transform(source).ast;
+  var program = ast.program;
+  console.log(
+    JSON.stringify(format(program), null, 2)
```
joshuaclayton: can we assign `null` and `2` to variables so it's easier to understand what they represent?

lib/cc/engine/analyzers/node.rb
```diff
@@ -0,0 +1,74 @@
...
+        def set_sexp_line(sexp)
+          sexp.line = @line
+        end
+
+        def properties_to_sexps
+          valid_properties.map do |key, value|
+            if value.is_a?(Array)
+              create_sexp(key.to_sym, *self.class.new(value, @file, @line).format)
+            elsif  value.is_a?(Hash)
+              create_sexp(key.to_sym, self.class.new(value, @file, @line).format)
```
joshuaclayton: could we batch these two together?

```ruby
create_sexp(key.to_sym, *Array(self.class.new(value, @file, @line).format))
```

BlakeWilliams:
I noticed that as well, that's something I was going to bring up and attempt to fix in another PR.

I think it's worth a discussion because it might make sense to allow users to provide an array of glob strings that we use to find the files so we could have something similar to:

```
engines:
  duplication:
    enabled: true
    languages:
      javascript:
        paths:
          -- **/*.js
          -- app/components/**/*.jsx
```

or some better format.

lib/cc/engine/analyzers/python/node.rb
```diff
@@ -1,69 +1,28 @@
...
-              @line = default
+              type
             end
           end
 
-          def has_line_number?
-            @node.is_a?(Hash) && @node["attributes"] && @node["attributes"]["lineno"]
+          def line_number
+            if @node.is_a?(Hash)
+              @node.fetch("attributes", {}).fetch("lineno", nil)
```
joshuaclayton: if the default is going to be `nil`, could we just use `["lineno"]`?

lib/cc/engine/analyzers/javascript/parser.js
BlakeWilliams: Oops, we don't even need those anymore since we don't care about pretty printing the output.

lib/cc/engine/analyzers/python/node.rb
BlakeWilliams: Good point, was just being explicit about it. I'll remove it.

lib/cc/engine/analyzers/node.rb
BlakeWilliams: I think I tried originally but since `Sexp` inherits from `Array` it will strip the outer s-expression giving us the following:

```ruby
s(:ctx, :Load)
```

Instead of the correct:

```ruby
s(:ctx, s(:Load))
```

fhwang:
Also, regarding file extensions, don't forget that some engines whitelist certain specific filenames as being relevant even without an extension. i.e., Rubocop reads `Rakefile`. I could imagine some languages where duplication analysis of these edge cases could be useful, like rake tasks.

Paths could be a good first step in handling that. If we let ourselves dream big, then reusable, language-specific whitelists could be useful in many places.

BlakeWilliams:
Agreed, I think that could evolve pretty far in the future with default, extensions (eg: [eslint](http://eslint.org/docs/user-guide/configuring#extending-configuration-files)), etc.

BlakeWilliams:
Any other feedback, or does this look good to merge?

GordonDiggs:
Cool - I like the paths approach

BlakeWilliams:
Awesome, I'll work on implementing what I pasted above in a new PR.

lib/cc/engine/analyzers/node.rb
```diff
@@ -0,0 +1,74 @@
...
+            self.class::SCRUB_PROPERTIES.include?(key) || value_empty
+          end
+        end
+
+        def set_default_line(default)
+          if line_number
+            @line = line_number
+          else
+            @line = default
+          end
```
mxie: Perhaps `@line = line_number || default`?

BlakeWilliams: Good catch, missed it when extracting things!

lib/cc/engine/analyzers/python/node.rb
BlakeWilliams: Turns out it raises sine `lineno` doesn't exist in `{}`. I think I'll stick to what's currently there.

lib/cc/engine/analyzers/node.rb
mxie: Could we somehow reuse the `format` method already defined? It seems to do similar things as the `self.class.stuff` argument here.

BlakeWilliams: We can't since format works on `type` and these are the properties that exist next to type.

eg:

```javascript
{ type: "class", name: "bar", block: { type: "Expression",}}
```

mxie:
Extensions concerns aside, this LGTM!

PR #9, opened 2015-10-15
Add configurable paths via engine config

BlakeWilliams:
This change lets users of the engine specify globs that the engine will
use to find files to analyze. We also provide sane defaults for each
language in case the user doesn't want to configure their paths.

Example config with this change:

```
engines:
  duplication:
    enabled: true
    config:
      languages:
        ruby:
          paths:
            - **/*.rb
            - **/*.rake
```

### DIFF
--- a/lib/cc/engine/analyzers/file_list.rb
+++ b/lib/cc/engine/analyzers/file_list.rb
@@ -2,23 +2,43 @@ module CC
   module Engine
     module Analyzers
       class FileList
-        def initialize(directory:, engine_config:, extension:)
+        def initialize(directory:, engine_config:, default_paths:, language:)
           @directory = directory
           @engine_config = engine_config
-          @extension = extension
+          @default_paths = default_paths
+          @language = language
         end
 
         def files
-          matching_files = Dir.glob("#{directory}/**/*.#{extension}").reject do |f|
-            File.directory?(f)
-          end
-
           matching_files - excluded_files
         end
 
         private
 
-        attr_reader :directory, :engine_config, :extension
+        attr_reader :directory, :engine_config, :default_paths, :language
+
+        def matching_files
+          paths.map do |glob|
+            Dir.glob("#{directory}/#{glob}").reject do |f|
+              File.directory?(f)
+            end
+          end.flatten
+        end
+
+        def paths
+          engine_paths || default_paths
+        end
+
+        def engine_paths
+          engine_language =  engine_config.
+            fetch("config", {}).
+            fetch("languages", {}).
+            fetch(language, {})
+
+          if engine_language.is_a?(Hash)
+            engine_language["paths"]
+          end
+        end
 
         def excluded_files
           excluded_paths.map { |path| Dir.glob("#{directory}/path") }.flatten
--- a/lib/cc/engine/analyzers/javascript/main.rb
+++ b/lib/cc/engine/analyzers/javascript/main.rb
@@ -9,6 +9,11 @@ module CC
     module Analyzers
       module Javascript
         class Main
+          LANGUAGE = "javascript"
+          DEFAULT_PATHS = [
+            "**/*.js",
+            "**/*.jsx"
+          ]
           DEFAULT_MASS_THRESHOLD = 40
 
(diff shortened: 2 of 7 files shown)


lib/cc/engine/analyzers/file_list.rb
```diff
@@ -2,23 +2,41 @@ module CC
...
           @engine_config = engine_config
-          @extension = extension
+          @default_paths = default_paths
+          @language = language
         end
 
         def files
-          matching_files = Dir.glob("#{directory}/**/*.#{extension}").reject do |f|
-            File.directory?(f)
+          matching_files = paths.map do |glob|
```
joshuaclayton: can this be extracted to a private method?

lib/cc/engine/analyzers/php/main.rb
```diff
@@ -7,6 +7,7 @@ module Engine
     module Analyzers
       module Php
         class Main
+          DEFAULT_PATHS = ["**/*.php"]
```
GordonDiggs: PHP also uses `.inc` and `.module` files - we support these on the Classic analysis

lib/cc/engine/analyzers/file_list.rb
```diff
@@ -2,23 +2,41 @@ module CC
...
+            Dir.glob("#{directory}/#{glob}").reject do |f|
+              File.directory?(f)
+            end
           end
 
-          matching_files - excluded_files
+          matching_files.flatten - excluded_files
```
joshuaclayton: if we extract matching files, I think we'll want to flatten there

lib/cc/engine/analyzers/file_list.rb
```diff
@@ -2,23 +2,41 @@ module CC
...
+        end
+
+        def engine_paths
+          engine_language =  engine_config.
+            fetch("config", {}).
+            fetch("languages", {}).
+            fetch(language, {})
+
+          if engine_language.is_a?(Hash)
+            engine_language.fetch("paths", nil)
```
joshuaclayton: thoughts on `engine_language["paths"]` instead of using `fetch`?

BlakeWilliams: :+1:

lib/cc/engine/analyzers/javascript/main.rb
```diff
@@ -42,7 +46,8 @@ def files
             ::CC::Engine::Analyzers::FileList.new(
               directory: directory,
               engine_config: engine_config,
-              extension: "js",
+              default_paths: DEFAULT_PATHS,
+              language: "javascript"
```
joshuaclayton: thoughts on extracting the language string to a constant as well?

lib/cc/engine/analyzers/php/main.rb
BlakeWilliams: Thanks, I had no idea. Adding that.

lib/cc/engine/analyzers/ruby/main.rb
```diff
@@ -7,6 +7,7 @@ module Engine
     module Analyzers
       module Ruby
         class Main
+          DEFAULT_PATHS = ["**/*.rb", "**/*.rake", "Rakefile"]
```
fhwang: For Ruby, you can find Rubocop's defaults here:

https://github.com/bbatsov/rubocop/blob/aa7d73992fcd1abbf3b57e56e6d9888a8e45d682/config/default.yml

BlakeWilliams: Does it make sense for duplication to check against all of those extensions?

fhwang: I haven't even used all those different Ruby libs, and, yeah, I bet some of them would be nonsensical to check for duplication. I think going through that list and using your own discretion is probably best.

GordonDiggs: Important to note that people can always configure them if need be

BlakeWilliams:
Good to :ship: ?

PR #10, opened 2015-10-15
Make config fetching safer

BlakeWilliams:
The current code makes a few assumptions about the format of your
`.codeclimate.yml` which can cause the engine to crash.

This changes the way we fetch the configuration to consider that not
only may the keys not be present, but they might also not always be the
type we expect.

### DIFF
--- a/lib/cc/engine/analyzers/file_list.rb
+++ b/lib/cc/engine/analyzers/file_list.rb
@@ -30,14 +30,21 @@ module CC
         end
 
         def engine_paths
-          engine_language =  engine_config.
-            fetch("config", {}).
-            fetch("languages", {}).
-            fetch(language, {})
-
-          if engine_language.is_a?(Hash)
-            engine_language["paths"]
+          if current_language.is_a?(Hash)
+            current_language["paths"]
           end
+        end
+
+        def current_language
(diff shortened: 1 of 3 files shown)


lib/cc/engine/analyzers/file_list.rb
```diff
@@ -30,16 +30,23 @@ def paths
...
-          if engine_language.is_a?(Hash)
-            engine_language["paths"]
+        def current_language
+          if engine_languages.is_a?(Hash)
+            @current_language ||= engine_languages.fetch(language, {})
           end
         end
 
+        def engine_languages
+          @engine_language ||= languages = engine_config.
```
GordonDiggs: what is `languages` used for?

BlakeWilliams: It's not, removing it.

GordonDiggs:
Is the purpose of this changeset to be backwards compatible?

BlakeWilliams:
Yes, and to prevent crashing when someone gives us an array of languages when they want to rely purely on the defaults.

GordonDiggs:
LGTM

BlakeWilliams:
Thanks for the review!

PR #13, opened 2015-10-19
Add engine_config class

BlakeWilliams:
This adds the `EngineConfig` class for normalizing the passed in engine
config in a single place instead of using `fetch` all over the codebase
with hardcoded paths.

This also fixes the path `mass_threshold` looks for when fetching the
config.

### DIFF
--- a/lib/cc/engine/analyzers/engine_config.rb
+++ b/lib/cc/engine/analyzers/engine_config.rb
@@ -1,0 +1,58 @@
+module CC
+  module Engine
+    module Analyzers
+      class EngineConfig
+        def initialize(hash)
+          @config = normalize(hash)
+        end
+
+        def exclude_paths
+          config.fetch("exclude_paths", [])
+        end
+
+        def languages
+          config.fetch("languages", {})
+        end
+
+        def mass_threshold_for(language)
+          fetch_language(language).fetch("mass_threshold", nil)
+        end
+
+        def paths_for(language)
+          fetch_language(language).fetch("paths", nil)
+        end
+
+        private
+
+        attr_reader :config
+
+        def fetch_language(language)
+          config.
+            fetch("languages", {}).
+            fetch(language, {})
+        end
+
+        def normalize(hash)
+          hash.tap do |config|
+            languages = config.fetch("config", {}).fetch("languages", {})
+            config["languages"] = build_language_config(languages)
+          end
+        end
+
+        def build_language_config(languages)
+          if languages.is_a?(Array)
+            languages.each_with_object({}) do |language, map|
+              map[language.downcase] = {}
+            end
+          elsif languages.is_a?(Hash)
+            languages.each_with_object({}) do |(key, value), map|
+              map[key.downcase] = value
+            end
+          else
+            {}
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/cc/engine/analyzers/file_list.rb
+++ b/lib/cc/engine/analyzers/file_list.rb
@@ -30,21 +30,7 @@ module CC
         end
 
         def engine_paths
-          if current_language.is_a?(Hash)
-            current_language["paths"]
-          end
-        end
-
-        def current_language
-          if engine_languages.is_a?(Hash)
-            @current_language ||= engine_languages.fetch(language, {})
-          end
-        end
-
-        def engine_languages
-          @engine_language ||= engine_config.
-            fetch("config", {}).
-            fetch("languages", {})
+          @engine_config.paths_for(language)
         end
 
         def excluded_files
@@ -52,7 +38,7 @@ module CC
         end
 
         def excluded_paths
-          Array(engine_config["exclude_paths"])
+          engine_config.exclude_paths
         end
       end
     end
--- a/lib/cc/engine/analyzers/javascript/main.rb
+++ b/lib/cc/engine/analyzers/javascript/main.rb
@@ -29,7 +29,7 @@ module CC
           end
 
           def mass_threshold
-            engine_config.fetch("config", {}).fetch("javascript", {}).fetch("mass_threshold", DEFAULT_MASS_THRESHOLD)
+            engine_config.mass_threshold_for(LANGUAGE) || DEFAULT_MASS_THRESHOLD
           end
 
           def base_points
--- a/lib/cc/engine/analyzers/php/main.rb
+++ b/lib/cc/engine/analyzers/php/main.rb
@@ -30,7 +30,7 @@ module CC
           end
 
           def mass_threshold
-            engine_config.fetch('config', {}).fetch('php', {}).fetch('mass_threshold', DEFAULT_MASS_THRESHOLD)
+            engine_config.mass_threshold_for(LANGUAGE) || DEFAULT_MASS_THRESHOLD
           end
 
           def base_points
--- a/lib/cc/engine/duplication.rb
+++ b/lib/cc/engine/duplication.rb
@@ -3,6 +3,7 @@ require 'cc/engine/analyzers/javascript/main'
 require 'cc/engine/analyzers/php/main'
 require 'cc/engine/analyzers/python/main'
 require 'cc/engine/analyzers/reporter'
+require 'cc/engine/analyzers/engine_config'
 require 'cc/engine/analyzers/sexp'
 require 'flay'
 require 'json'
@@ -10,7 +11,7 @@ require 'json'
 module CC
   module Engine
     class Duplication
-      DEFAULT_LANGUAGE = ['ruby'].freeze
+      DEFAULT_LANGUAGE = 'ruby'.freeze
       LANGUAGES = {
         "ruby"       => ::CC::Engine::Analyzers::Ruby::Main,
         "javascript" => ::CC::Engine::Analyzers::Javascript::Main,
@@ -20,7 +21,7 @@ module CC
 
       def initialize(directory:, engine_config:, io:)
         @directory = directory
-        @engine_config = engine_config || {}
+        @engine_config = CC::Engine::Analyzers::EngineConfig.new(engine_config || {})
         @io = io
       end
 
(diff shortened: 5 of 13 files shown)


BlakeWilliams:
cc @codeclimate/review

lib/cc/engine/analyzers/engine_config.rb
```diff
@@ -0,0 +1,49 @@
...
+
+        def build_language_config(config)
+          languages = config.fetch("languages", {})
+
+          if languages.is_a?(Array)
+            languages.each_with_object({}) do |language, map|
+              map[language.downcase] = {}
+            end
+          elsif languages.is_a?(Hash)
+            languages.each_with_object({}) do |language, map|
```
mxie: What if we did something like `|(key, value), map|` here instead of doing the deconstructing in the line below?

lib/cc/engine/analyzers/engine_config.rb
```diff
@@ -0,0 +1,49 @@
...
+              key, value = language[0], language[1]
+              map[key.downcase] = value
```
mxie: Could we default to `{}` here?

lib/cc/engine/analyzers/engine_config.rb
```diff
@@ -0,0 +1,49 @@
...
+      class EngineConfig
+        attr_reader :config
+
+        def initialize(hash)
+          normalize(hash)
+        end
+
+        def language(language)
+          config.
+            fetch("languages", {}).
```
mxie: If we're good at providing the default values in `build_language_config`, could we just not do this `fetch` stuff?

BlakeWilliams: I think we can. I'm probably being too paranoid about getting weird user input.

lib/cc/engine/analyzers/engine_config.rb
BlakeWilliams: Hah, I did `[]` last time. Too much Elixir. :smile:

lib/cc/engine/analyzers/engine_config.rb
```diff
@@ -0,0 +1,49 @@
...
+          config.fetch("languages", {})
+        end
+
+        private
+
+        attr_reader :hash
+
+        def normalize(hash)
+          @config = hash.fetch("config", {})
+          @config["languages"] = build_language_config(@config)
```
wfleming: Minor nit, but it's a bit weird that we have a mutating method that gets passed in an instance member, mutates that instance member (when it exists), and then overwrites part of the passed-in instance member with the return value.

BlakeWilliams: Good point, I'll just make it mutate in the private method.

lib/cc/engine/analyzers/engine_config.rb
wfleming: I think paranoia is probably appropriate here.

That said, out of curiosity, what is the engine's behavior if no languages are specified?

lib/cc/engine/analyzers/engine_config.rb
BlakeWilliams: Actually, it's easier just to return the new config instead of making the method mutate config.

lib/cc/engine/analyzers/engine_config.rb
BlakeWilliams: We should use the `DEFAULT_LANGUAGE`, which is Ruby.

https://github.com/codeclimate/codeclimate-duplication/blob/bmw-engine-config-refactor/lib/cc/engine/duplication.rb#L49-L53

lib/cc/engine/duplication.rb
```diff
@@ -43,14 +44,12 @@ def languages_to_analyze
...
       def languages
-        config_languages = engine_config.
-          fetch("config", {}).
-          fetch("languages", DEFAULT_LANGUAGE)
+        languages = engine_config.languages
 
-        if config_languages.is_a?(Hash)
-          config_languages.keys
+        if languages.empty?
+          DEFAULT_LANGUAGE
```
joshuaclayton: should this be `DEFAULT_LANGUAGES`?

joshuaclayton: Or, what about having `DEFAULT_LANGUAGE` just return `"ruby"` instead of an array?

BlakeWilliams: I like that, I'll wrap it in `Array`

lib/cc/engine/analyzers/engine_config.rb
```diff
@@ -0,0 +1,48 @@
+module CC
+  module Engine
+    module Analyzers
+      class EngineConfig
+        attr_reader :config
```
joshuaclayton: does this need to be public?

lib/cc/engine/analyzers/engine_config.rb
```diff
@@ -0,0 +1,48 @@
...
+
+        def initialize(hash)
+          normalize(hash)
```
joshuaclayton: Thoughts on normalization happening in another class? I don't think this is high priority, but it seems like pulling data out of YAML is a separate concern from loading up configuration.

lib/cc/engine/analyzers/engine_config.rb
BlakeWilliams: Not anymore, good catch,

lib/cc/engine/analyzers/engine_config.rb
```diff
@@ -0,0 +1,48 @@
...
+        def languages
+          config.fetch("languages", {})
+        end
+
+        private
+
+        attr_reader :hash
+
+        def normalize(hash)
+          @config = hash.fetch("config", {})
```
joshuaclayton: thoughts on moving this to the constructor to avoid mutation?

lib/cc/engine/analyzers/engine_config.rb
```diff
@@ -0,0 +1,48 @@
...
+          if languages.is_a?(Array)
+            languages.each_with_object({}) do |language, map|
+              map[language.downcase] = {}
+            end
+          elsif languages.is_a?(Hash)
+            languages.each_with_object({}) do |(key, value), map|
+              map[key.downcase] = value
+            end
+          else
+            hash.languages = {}
```
joshuaclayton: Does this branch of code get executed?

BlakeWilliams: It could if the user gave invalid input.

lib/cc/engine/analyzers/engine_config.rb
BlakeWilliams: :+1:

lib/cc/engine/analyzers/file_list.rb
```diff
@@ -30,21 +30,7 @@ def paths
...
-          if engine_languages.is_a?(Hash)
-            @current_language ||= engine_languages.fetch(language, {})
-          end
-        end
-
-        def engine_languages
-          @engine_language ||= engine_config.
-            fetch("config", {}).
-            fetch("languages", {})
+          @engine_config.language(language).fetch("paths", nil)
```
joshuaclayton: instead of other areas of the codebase being concerned with the structure of this data in `engine_config`, what are your thoughts on creating public methods to get at this data so `EngineConfig` is solely responsible for all of its data?

lib/cc/engine/analyzers/file_list.rb
```diff
@@ -52,7 +38,7 @@ def excluded_files
         end
 
         def excluded_paths
-          Array(engine_config["exclude_paths"])
+          Array(engine_config.config["exclude_paths"])
```
joshuaclayton: Same thing here re: `engine_config` owning its data, and providing methods to interact with (sanitized) information?

lib/cc/engine/analyzers/php/main.rb
```diff
@@ -30,7 +30,9 @@ def run
           end
 
           def mass_threshold
-            engine_config.fetch('config', {}).fetch('php', {}).fetch('mass_threshold', DEFAULT_MASS_THRESHOLD)
+            engine_config.
+              language(LANGUAGE).
+              fetch("mass_threshold", DEFAULT_MASS_THRESHOLD)
```
joshuaclayton: same here - e.g. `engine_config.language.mass_threshold || DEFAULT_MASS_THRESHOLD`

BlakeWilliams: I'm a fan of that.

lib/cc/engine/analyzers/engine_config.rb
joshuaclayton: @BlakeWilliams where does `hash` get assigned? I see it's an `attr_reader`, but it's unclear where it's actually set.

BlakeWilliams: Oh, that was an old branch of code, it isn't getting hit there. I'll add a test to cover that case.

BlakeWilliams:
Good to go?

lib/cc/engine/analyzers/javascript/main.rb
```diff
@@ -29,7 +29,7 @@ def run
           end
 
           def mass_threshold
```
wfleming: Maybe now is not the time, but it seems like these per-language `main` classes are a good case for inheritance?

wfleming:
One refactor suggestion, but maybe it should be saved for later. LGTM.

lib/cc/engine/analyzers/javascript/main.rb
BlakeWilliams: Yep, I started thinking the same thing and was going to look into it after this PR is good to go.

BlakeWilliams:
Yeah, I think it's a separate concern so I'll handle that in another PR.

BlakeWilliams:
Thanks for the reviews!

PR #14, opened 2015-10-19
Extract base analyzer class

BlakeWilliams:
This extracts common code from the analyzers into the `AnalyzerBase`
class.

The only requirements for an analyzer now are having the following
constants:

* `DEFAULT_PATHS` - default glob of files to analyze
* `LANGUAGE` - The analyzers language name as a string
* `DEFAULT_MASS_THRESHOLD` - Default mass we should tell Flay to use
* `BASE_POINTS` - Amount of points to multiply by mass

and the following method(s):

* `process_file(path)` - Returns s-expressions based on the file content
  of `path`.

### DIFF
--- a/lib/cc/engine/analyzers/analyzer_base.rb
+++ b/lib/cc/engine/analyzers/analyzer_base.rb
@@ -1,0 +1,43 @@
+module CC
+  module Engine
+    module Analyzers
+      class Base
+        def initialize(engine_config:, directory:)
+          @engine_config = engine_config
+          @directory = directory
+        end
+
+        def run
+          files.map do |file|
+            process_file(file)
+          end
+        end
+
+        def mass_threshold
+          engine_config.mass_threshold_for(self.class::LANGUAGE) || self.class::DEFAULT_MASS_THRESHOLD
+        end
+
+        def base_points
+          self.class::BASE_POINTS
+        end
+
+        private
+
+        attr_reader :engine_config, :directory
+
+        def process_file(path)
+          raise NoMethodError.new("Subclass must implement `process_file`")
+        end
+
+        def files
+          ::CC::Engine::Analyzers::FileList.new(
+            directory: directory,
+            engine_config: engine_config,
+            default_paths: self.class::DEFAULT_PATHS,
+            language: self.class::LANGUAGE
+          ).files
+        end
+      end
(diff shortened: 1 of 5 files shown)


BlakeWilliams:
Should be a small one @codeclimate/review

lib/cc/engine/analyzers/analyzer_base.rb
```diff
@@ -0,0 +1,43 @@
+module CC
+  module Engine
+    module Analyzers
+      class AnalyzerBase
```
GordonDiggs: I think it might be a bit more idiomatic if this were named `CC::Engine::Analyzers::Base`. What do you think?

BlakeWilliams: I was thinking about it but marginally decided against it. I'll change it to be just `Base`.

lib/cc/engine/analyzers/analyzer_base.rb
```diff
@@ -0,0 +1,43 @@
...
+        def base_points
+          self.class::BASE_POINTS
+        end
+
+        private
+
+        attr_reader :engine_config, :directory
+
+        def process_file(path)
+          raise "Subclass must implement `process_file`"
```
GordonDiggs: Can you raise a [NotImplementedError](http://ruby-doc.org/core-2.2.0/NotImplementedError.html) here?

BlakeWilliams: I don't think we want to raise a `NotImplementedError` since it's usage is supposed to be when a feature isn't implemented on the current platform. Thoughts?

BlakeWilliams: Maybe raise http://ruby-doc.org/core-2.2.0/NoMethodError.html ?

GordonDiggs: Ah interesting. We use it elsewhere as a convention for this abstract class purpose. I think `NoMethodError` would be incorrect. Maybe just best to leave it as-is

GordonDiggs:
Minor nits but otherwise LGTM. :clap: love the red

BlakeWilliams:
Thanks for the review!

PR #15, opened 2015-10-22
Rescue and skip file when `process_file` fails

BlakeWilliams:
This adds a begin/rescue/end block around `process_file(file)` calls so
when the analyze fails to analyze a file it is skipped instead of
crashing the entire engine.

Is there any type of logging we can do that would be valuable in the rescue block?

### DIFF
--- a/lib/cc/engine/analyzers/analyzer_base.rb
+++ b/lib/cc/engine/analyzers/analyzer_base.rb
@@ -9,7 +9,12 @@ module CC
 
         def run
           files.map do |file|
-            process_file(file)
+            begin
+              process_file(file)
+            rescue => e
+              $stderr.puts "Skipping file #{file} due to exception"
+              $stderr.puts "(#{ex.class}) #{ex.message} #{ex.backtrace.join("\n")}"
+            end
           end
         end
 


pbrisbin:
Engines can emit to `stderr`. Eventually we'll have a richer "Warning" object that can be emitted on `stdout`, but we haven't built handling for it yet.

/cc @mrb

mrb:
@pbrisbin Does this stderr stuff get logged anywhere and/or is there a way to surface it as something we can track?

pbrisbin:
In the CLI, `stderr` is shown if the engine exits non-zero.

On dot-com, stderr is recorded always and is visible in the case of an errored snapshot step.

mrb:
@pbrisbin Okay, makes sense, thanks.

BlakeWilliams:
Do we want to log "Skipped #{file}" to stderr then?

pbrisbin:
Since you've got a bare `rescue`, I'd probably log the exception details exactly.

BlakeWilliams:
Added.

pbrisbin:
Maybe not pushed?

BlakeWilliams:
Oops! My mistake.

lib/cc/engine/analyzers/analyzer_base.rb
```diff
@@ -9,7 +9,11 @@ def initialize(engine_config:, directory:)
 
         def run
           files.map do |file|
-            process_file(file)
+            begin
+              process_file(file)
+            rescue => e
+              $stderr.puts e
```
pbrisbin: I'm not sure what this will look like. I might get a touch fancier / more explicit:

```rb
$stderr.puts "Skipping file #{file} due to exception"
$stderr.puts "(#{ex.class}) #{ex.message} #{ex.backtrace.join("\n")}"
```

BlakeWilliams:
```
Skipping file ../paperclip/lib/paperclip/attachment.rb due to exception
(Racc::ParseError) ../paperclip/lib/paperclip/attachment.rb:7 :: parse error on value 1 (tINTEGER) /Users/Blake/.rbenv/versions/2.2.2/lib/ruby/2.2.0/racc/parser.rb:527:in `on_error'
/Users/Blake/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/ruby_parser-3.7.1/lib/ruby_parser_extras.rb:1124:in `on_error'
/Users/Blake/.rbenv/versions/2.2.2/lib/ruby/2.2.0/racc/parser.rb:258:in `_racc_do_parse_c'
/Users/Blake/.rbenv/versions/2.2.2/lib/ruby/2.2.0/racc/parser.rb:258:in `do_parse'
/Users/Blake/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/ruby_parser-3.7.1/lib/ruby_parser_extras.rb:1036:in `block in process'
/Users/Blake/.rbenv/versions/2.2.2/lib/ruby/2.2.0/timeout.rb:89:in `block in timeout'
/Users/Blake/.rbenv/versions/2.2.2/lib/ruby/2.2.0/timeout.rb:34:in `block in catch'
/Users/Blake/.rbenv/versions/2.2.2/lib/ruby/2.2.0/timeout.rb:34:in `catch'
/Users/Blake/.rbenv/versions/2.2.2/lib/ruby/2.2.0/timeout.rb:34:in `catch'
/Users/Blake/.rbenv/versions/2.2.2/lib/ruby/2.2.0/timeout.rb:104:in `timeout'
/Users/Blake/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/ruby_parser-3.7.1/lib/ruby_parser_extras.rb:1024:in `process'
/Users/Blake/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/ruby_parser-3.7.1/lib/ruby_parser_extras.rb:1362:in `block in process'
/Users/Blake/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/ruby_parser-3.7.1/lib/ruby_parser_extras.rb:1360:in `each'
/Users/Blake/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/ruby_parser-3.7.1/lib/ruby_parser_extras.rb:1360:in `process'
/Users/Blake/Documents/Projects/codeclimate-duplication/lib/cc/engine/analyzers/ruby/main.rb:29:in `process_file'
/Users/Blake/Documents/Projects/codeclimate-duplication/lib/cc/engine/analyzers/analyzer_base.rb:13:in `block in run'
/Users/Blake/Documents/Projects/codeclimate-duplication/lib/cc/engine/analyzers/analyzer_base.rb:11:in `map'
/Users/Blake/Documents/Projects/codeclimate-duplication/lib/cc/engine/analyzers/analyzer_base.rb:11:in `run'
/Users/Blake/Documents/Projects/codeclimate-duplication/lib/cc/engine/analyzers/reporter.rb:15:in `run'
/Users/Blake/Documents/Projects/codeclimate-duplication/lib/cc/engine/duplication.rb:32:in `block in run'
/Users/Blake/Documents/Projects/codeclimate-duplication/lib/cc/engine/duplication.rb:29:in `each'
/Users/Blake/Documents/Projects/codeclimate-duplication/lib/cc/engine/duplication.rb:29:in `run'
./bin/duplication:16:in `<main>'
```

BlakeWilliams: The previous comment is what your output looks like.

This is what it currently looks like:

```
../paperclip/lib/paperclip/attachment.rb:7 :: parse error on value 1 (tINTEGER)
```

pbrisbin: I'm fine with the simpler `puts e` syntax and omitting the backtrace. I think you should keep the "skipping file" line.

WDYT @mrb ?

BlakeWilliams: Agreed, "skipping file" is a good addition.

mrb: @pbrisbin @BlakeWilliams Hmmmmmm - maybe SOME backtrace? Seems helpful!

BlakeWilliams: I think we should stick with the full backtrace since we don't know how much we need to display for the information to be useful?

BlakeWilliams:
I added the full trace and the skip line. Any more feedback before I merge?

lib/cc/engine/analyzers/analyzer_base.rb
```diff
@@ -9,7 +9,12 @@ def initialize(engine_config:, directory:)
 
         def run
           files.map do |file|
-            process_file(file)
+            begin
+              process_file(file)
+            rescue => e
```
pbrisbin: Should this go in `#process_file` so you can use an implicit `begin`? I usually take explicit `begin`s as a smell pointing at ExtractMethod -- and you've already done that, so might as well take advantage.

pbrisbin:
:lipstick: nit -- LGTM besides.

lib/cc/engine/analyzers/analyzer_base.rb
BlakeWilliams: The reason I didn't do it in `process_file` is because each language defines that method themselves.

I don't think it's worth duplicating this in each analyzer. Thoughts?

pbrisbin: Gotcha, that's fine with me then

PR #16, opened 2015-10-26
Allow empty language keys in config

BlakeWilliams:
This allows `.codeclimate.yml` language keys to have empty values and
still be valid.

eg:

```
engines:
  duplication:
    enabled: true
    config:
      languages:
        javascript:
```


noahd1:
LGTM

PR #17, opened 2015-10-26
Mimic descriptions emitted by classic

jpignata:
Previously Code Climate would show a description that indicated how many
instances of duplication were found within a given issue. This seems
more intelligble than what this engine does by pointing to data from the
s-expression reported by flay.


BlakeWilliams:
Looks good to me!

PR #19, opened 2015-10-27
Always return hash for `fetch_language`

BlakeWilliams:
Fetch language is used to grab the configuration for an individual
language and is expected to be a hash. This change makes
`fetch_language` return an empty hash if language isn't already a hash.

### DIFF
--- a/spec/cc/engine/analyzers/engine_config_spec.rb
+++ b/spec/cc/engine/analyzers/engine_config_spec.rb
@@ -1,5 +1,6 @@
 require "spec_helper"
 require "cc/engine/analyzers/engine_config"
+require "cc/engine/analyzers/ruby/main"
 
 module CC::Engine::Analyzers
   describe EngineConfig do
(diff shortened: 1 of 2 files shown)


spec/cc/engine/analyzers/engine_config_spec.rb
```diff
@@ -84,6 +85,18 @@ module CC::Engine::Analyzers
...
+      it "returns nil when language is empty" do
+        engine_config = EngineConfig.new({
+          "config" => {
+            "languages" => {
+              "ruby" => "",
+            }
+          }
+        })
+
+        assert_equal engine_config.mass_threshold_for("ruby"), nil
```
GordonDiggs: I'm confused by this spec. Above you change `paths_for` but this spec tests `mass_threshold_for`. Should there be another spec?

BlakeWilliams: `paths_for` was changed to be safe in a previous PR, but this PR makes that change unnecessary since `fetch_language` should now always return a safe value.

I wrote this one to get the failure, then changed `paths_for` to work as well.

GordonDiggs:
LGTM

BlakeWilliams:
Thanks!

PR #20, opened 2015-10-27
Convert configured mass threshold to integer

BlakeWilliams:
Currently a configured mass threshold will be returned as a string when
flay expects an integer when it tries to compare mass to mass threshold.

This also renames the assertion since the previous title was incorrect.

### DIFF
--- a/lib/cc/engine/analyzers/engine_config.rb
+++ b/lib/cc/engine/analyzers/engine_config.rb
@@ -15,7 +15,11 @@ module CC
         end
 
         def mass_threshold_for(language)
-          fetch_language(language).fetch("mass_threshold", nil)
+          threshold = fetch_language(language).fetch("mass_threshold", nil)
+
+          if threshold
+            threshold.to_i
+          end
         end
 
         def paths_for(language)
--- a/spec/cc/engine/analyzers/engine_config_spec.rb
+++ b/spec/cc/engine/analyzers/engine_config_spec.rb
@@ -72,12 +72,12 @@ module CC::Engine::Analyzers
     end
 
     describe "mass_threshold_for" do
-      it "returns empty hash if language is not present" do
+      it "returns configured mass threshold as integer" do
         engine_config = EngineConfig.new({
           "config" => {
             "languages" => {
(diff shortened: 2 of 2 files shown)


lib/cc/engine/analyzers/engine_config.rb
```diff
@@ -15,7 +15,11 @@ def languages
         end
 
         def mass_threshold_for(language)
-          fetch_language(language).fetch("mass_threshold", nil)
+          threshold = fetch_language(language).fetch("mass_threshold", nil)
```
GordonDiggs: What happens when this returns nil? Do we default it somewhere?

BlakeWilliams: Yep, https://github.com/codeclimate/codeclimate-duplication/blob/master/lib/cc/engine/analyzers/analyzer_base.rb#L22

jpignata:
Why isn't the YAML parsing properly casting this value to an integer?

jpignata:
Do we have an upstream bug where we're incorrectly casting the value from an integer to a string in the YAML -> JSON translation?

BlakeWilliams:
That may be it, by the time it gets to the `mass_threshold_for` it's a string.

```
    config:
      languages:
        ruby:
          mass_threshold: 15
```

Returns `"15"` when fetching it.

BlakeWilliams:
For reference, when running `YAML.load("foo: 15")` you get `15` as an integer.

pbrisbin:
I would have to guess it's this logic that introduced an incorrect cast within the "arbitrary" `config` node: https://github.com/codeclimate/codeclimate-yaml/commit/5f6323f4762ccf14b1c823a5e11195a3552f1e54

jpignata:
Being defensive here is probably right regardless so :+1: from me on this pull. I'll log a card to look at the upstream bug so it doesn't get lost.

BlakeWilliams:
I also think being defensive is the best approach since it will work even if they pass `mass_threshold: "15"`.

GordonDiggs:
:+1: as well

PR #23, opened 2015-10-27
Fix exclude paths

BlakeWilliams:
The line that globs exclude paths had the word `path` hardcoded and
was never actually excluding the paths properly.

This adds a spec and interpolates the path properly.

I'm surprised this made it through as is throughout the refactorings.

### DIFF
--- a/lib/cc/engine/analyzers/file_list.rb
+++ b/lib/cc/engine/analyzers/file_list.rb
@@ -34,7 +34,7 @@ module CC
         end
 
         def excluded_files
-          excluded_paths.map { |path| Dir.glob("#{directory}/path") }.flatten
+          @_excluded_files ||= excluded_paths.map { |path| Dir.glob("#{directory}/#{path}") }.flatten
         end
 
         def excluded_paths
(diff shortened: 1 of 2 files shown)


dblandin:
LGTM

ABaldwinHunter:
Too late but yes :+1:

lib/cc/engine/analyzers/file_list.rb
```diff
@@ -34,7 +34,7 @@ def engine_paths
         end
 
         def excluded_files
-          excluded_paths.map { |path| Dir.glob("#{directory}/path") }.flatten
+          excluded_paths.map { |path| Dir.glob("#{directory}/#{path}") }.flatten
```
GordonDiggs: We should probably memoize this - the Dir.glob call can be expensive

BlakeWilliams: I think this only ever gets called once, but I can memoize it just to be safe.

GordonDiggs:
LGTM save for memoization

PR #24, opened 2015-10-27
Warn on stderr and raise an exception if no languages are enabled

dblandin:
Instead of picking a default language, the change in this PR will warn on stderr and raise an exception when no languages are enabled for duplication analysis.

### DIFF
--- a/lib/cc/engine/duplication.rb
+++ b/lib/cc/engine/duplication.rb
@@ -11,7 +11,8 @@ require 'json'
 module CC
   module Engine
     class Duplication
-      DEFAULT_LANGUAGE = 'ruby'.freeze
+      EmptyLanguagesError = Class.new(StandardError)
+
       LANGUAGES = {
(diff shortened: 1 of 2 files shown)


lib/cc/engine/duplication.rb
```diff
@@ -31,6 +32,10 @@ def run
           reporter = CC::Engine::Analyzers::Reporter.new(directory, engine, io)
           reporter.run
         end
+
+      rescue EmptyLanguagesError => ex
+        $stderr.puts ex.message
```
GordonDiggs: What do you think of doing the `$stderr.puts ...` below (inside of `languages.empty?`)? That way you wouldn't have to rescue and re-raise

dblandin: Sounds good to me :+1:

GordonDiggs:
LGTM

PR #25, opened 2015-10-27
Ensure value is always defined before access

jpignata:
In parsing some JSX from relax/relax, we noticed the parser was choking
on trying to access attributes of the loc attribute of a given node. It
looks like we were checking for the presence of value in every other
branch so this change hoists that check up around the processing.


GordonDiggs:
LGTM

PR #27, opened 2015-10-28
Raise the default Ruby mass threshold to 18

BlakeWilliams:
Currently the mass threshold is too sensitive and allows blocks of code
that shouldn't be compared through. This increases the threshold to 18
to help the engine feel less sensitive.


GordonDiggs:
LGTM

PR #28, opened 2015-10-29
chdir into code directory and use relative paths

BlakeWilliams:
Currently the engine uses absolute paths to `/code` when working with
files to analyze.

Now instead of using absolute paths to the files to analyze we `chdir`
to the given code directory and read files relative to the current path.
This will make the migration to use `include_paths` from `exclude_paths`
easier.

### DIFF
--- a/lib/cc/engine/analyzers/php/main.rb
+++ b/lib/cc/engine/analyzers/php/main.rb
@@ -20,7 +20,7 @@ module CC
           private
 
           def process_file(path)
-            code = File.read(path)
+            code = File.binread(path)
             parser = php_parser.new(code, path).parse
             syntax_tree = parser.syntax_tree
 
--- a/spec/cc/engine/analyzers/php/main_spec.rb
+++ b/spec/cc/engine/analyzers/php/main_spec.rb
@@ -2,12 +2,20 @@ require 'spec_helper'
 require 'cc/engine/analyzers/php/main'
 require 'cc/engine/analyzers/reporter'
 require 'cc/engine/analyzers/engine_config'
+require 'cc/engine/analyzers/file_list'
 require 'flay'
 require 'tmpdir'
 
 module CC::Engine::Analyzers::Php
   describe Main do
-    before { @code = Dir.mktmpdir }
+    before do
+      @code = Dir.mktmpdir
+      Dir.chdir(@code)
+    end
+
+    after do
+      FileUtils.rm_rf(@code)
+    end
 
     describe "#run" do
       it "prints an issue" do
@@ -41,8 +49,8 @@ module CC::Engine::Analyzers::Php
       def run_engine(config = nil)
         io = StringIO.new
 
-        engine = ::CC::Engine::Analyzers::Php::Main.new(directory: @code, engine_config: config)
-        reporter = ::CC::Engine::Analyzers::Reporter.new(@code, engine, io)
+        engine = ::CC::Engine::Analyzers::Php::Main.new(engine_config: config)
+        reporter = ::CC::Engine::Analyzers::Reporter.new(engine, io)
 
         reporter.run
 
--- a/lib/cc/engine/analyzers/analyzer_base.rb
+++ b/lib/cc/engine/analyzers/analyzer_base.rb
@@ -2,9 +2,8 @@ module CC
   module Engine
     module Analyzers
       class Base
-        def initialize(engine_config:, directory:)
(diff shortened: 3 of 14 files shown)


lib/cc/engine/analyzers/php/main.rb
```diff
@@ -20,7 +20,7 @@ class Main < CC::Engine::Analyzers::Base
           private
 
           def process_file(path)
-            code = File.read(path)
+            code = File.binread(path)
```
wfleming: This doesn't seem related to the dir change. Is there something in PHP source or how we parse it that prompted this?

BlakeWilliams: I was debugging and changed it just to see if this had anything to do with it, it's something we should probably be doing anyways since the other analyzers are using `binread` anyways.

BlakeWilliams: It's such a minor change I thought I'd keep it in since it's probably not worth its own PR.

wfleming: Makes sense to me. I didn't check the others to see if they were using `read` or `binread`.

spec/cc/engine/analyzers/php/main_spec.rb
```diff
@@ -2,12 +2,17 @@
...
+require 'cc/engine/analyzers/file_list'
 require 'flay'
 require 'tmpdir'
 
 module CC::Engine::Analyzers::Php
   describe Main do
-    before { @code = Dir.mktmpdir }
+    before do
+      @code = Dir.mktmpdir
+      Dir.chdir(@code)
```
wfleming: It is likely you will eventually end up being sad because some other spec that deals with relative paths runs with a different CWD than it expects to because of this.

I would suggest making this an `around` call that uses `chdir's` block syntax & yields to the spec.

BlakeWilliams: I was thinking that too initially, but Minitest doesn't have `around` and it errors out pretty quickly.

wfleming: Also, for neatness' sake, can we destroy the `@code` dirs after the spec runs? Is there a good reason to keep them around?

BlakeWilliams: I'd be all for migrating to RSpec though, if that's something we want to do so we can have nice things like `around`.

BlakeWilliams: Since it's a tmpdir, it *should* get wiped out on its own later.

pbrisbin: For now there's a `minitest/around` library (I think that's the name) that we've used in other places. I'm in favor of anyone that wants to migrate a codebase to RSpec though. I repeatedly headdesk over various aspects of minitest when I find myself in such a codebase.

wfleming: Well, you can simulate `around` with `before` & `after`, and `chdir` back to the correct cwd after the run, right?

I'm thinking about things like code coverage we might eventually integrate, randomized test order (which I wish minitest did out of the box), etc. Basically, unexpected working directory is one of those things that will eventually bite you, and is a pain to debug when it does. We should avoid it if we can.

BlakeWilliams: Yeah, I think the migration to RSpec would be more valuable than pulling in `minitest/around` here since this feels a lot like RSpec but has a lot of quirks that catch you by surprise.

Can we keep this as is for now and in the next PR I'll migrate to RSpec so we can have `around` and fix this issue?

wfleming: `/tmp` files generally aren't wiped out locally until you restart your system (and not always even then): if you create a file in `/tmp` and don't need it anymore, I think it's good practice to delete it yourself.

BlakeWilliams: That's fair, on OS X they're deleted about every 3 days at least. I do agree that it's probably a good practice though. I'll add that in.

BlakeWilliams:
Added the `after` hooks to cleanup after tmp_dirs. Does this look good to go? I plan on a follow up PR migrating to RSpec and that should solve the issue with `around` and `chdir`.

BlakeWilliams:
Thanks for the review!

PR #29, opened 2015-10-29
Use git to fetch composer dependencies

BlakeWilliams:
Composer occasionally fails the build on CircleCI because it is hitting
Github API rate limits.

According to https://circleci.com/docs/composer-api-rate-limit we can
use the `--prefer-source` and `--no-interaction` flags when calling
`composer install` to get around this.


wfleming:
:+1:

PR #30, opened 2015-10-29
Migrate from Minitest to RSpec

BlakeWilliams:
This migrates the test suite to RSpec since minitest is missing basic
things like `expect` syntax, `around` blocks, and a few other features.

While Minitest has plugins that provide these features, RSpec comes with
these features out of the box and more. Most of us are also seemingly
more familiar with RSpec.

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,0 +1,1 @@
+--require spec_helper
--- a/spec/cc/engine/analyzers/php/main_spec.rb
+++ b/spec/cc/engine/analyzers/php/main_spec.rb
@@ -1,4 +1,3 @@
-require 'spec_helper'
 require 'cc/engine/analyzers/php/main'
 require 'cc/engine/analyzers/reporter'
 require 'cc/engine/analyzers/engine_config'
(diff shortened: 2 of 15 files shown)


.rspec
```diff
@@ -0,0 +1,2 @@
+--color
+--require spec_helper
```
wfleming: :+1: we should do this everywhere.

spec/cc/engine/analyzers/php/main_spec.rb
```diff
@@ -1,78 +0,0 @@
-require 'spec_helper'
```
wfleming: Errr...intentional that this was just dropped entirely & not migrated?

BlakeWilliams: Oops, missed PHP.

.rspec
pbrisbin: I could get behind the implicit `--require spec_helper`, but I'm not sure I want to force `--color` on everyone, sounds like something to put in `~/.rspec` if that's how you roll.

wfleming: Yeah, I was referring specifically to the `--require`. Color does seem more user-preference.

BlakeWilliams: Agreed, this is just what RSpec generates by default. I'll remove color and remove `require 'spec_helper'` from existing specs.

spec/cc/engine/analyzers/php/main_spec.rb
BlakeWilliams: Added it back in.

wfleming:
:+1:

BlakeWilliams:
Thanks for the reviews!

PR #31, opened 2015-10-29
Replace exclude_paths with include_paths

BlakeWilliams:
`exclude_paths` has been deprecated in favor of `include_paths`.

Since each language allows you to specify its own paths to look for
duplications we have to build an array of files that the `include_paths`
specifies and then intersect those files with the files that match the
languages specified paths.

### DIFF
--- a/lib/cc/engine/analyzers/file_list.rb
+++ b/lib/cc/engine/analyzers/file_list.rb
@@ -1,3 +1,5 @@
+require "pathname"
+
 module CC
   module Engine
     module Analyzers
@@ -9,7 +11,7 @@ module CC
         end
 
         def files
-          matching_files - excluded_files
+          Array(matching_files) & Array(included_files)
         end
 
         private
@@ -18,8 +20,8 @@ module CC
 
         def matching_files
           paths.map do |glob|
-            Dir.glob("./#{glob}").reject do |f|
-              File.directory?(f)
+            Dir.glob("./#{glob}").reject do |path|
+              File.directory?(path)
             end
           end.flatten
         end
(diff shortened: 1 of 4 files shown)


lib/cc/engine/analyzers/file_list.rb
```diff
@@ -9,7 +11,7 @@ def initialize(engine_config:, default_paths:, language:)
         end
 
         def files
-          matching_files - excluded_files
+          Array(matching_files) & Array(included_files)
```
GordonDiggs: `&` is for intersection (in both A and B) - do you want union here (`|`)?

lib/cc/engine/analyzers/file_list.rb
```diff
@@ -18,8 +20,8 @@ def files
 
         def matching_files
           paths.map do |glob|
-            Dir.glob("./#{glob}").reject do |f|
-              File.directory?(f)
+            Dir.glob("./#{glob}").reject do |path|
+              File.directory?(path)
```
GordonDiggs: This will include symlinks, is that intentional?

GordonDiggs: To clarify, with this line, the final value will include symlinks

lib/cc/engine/analyzers/file_list.rb
BlakeWilliams: No, we only want files that are in both arrays since `include_paths` is created from the `.codeclimate.yml`s `exclude_paths` and this engine allows you to specify your own paths for each language. So you want to only take the matching files from the total available pool (`include_paths`) and get the files that match from that pool (matching_files here).

GordonDiggs: Got it!

lib/cc/engine/analyzers/file_list.rb
BlakeWilliams: It already was, so we should probably keep it that way. If we're running into symlinks with engines we should look at fixing that somewhere upstream.

lib/cc/engine/analyzers/file_list.rb
```diff
@@ -32,12 +34,33 @@ def engine_paths
...
+          include_paths.
+            map { |path| make_relative(path) }.
+            map { |path| collect_files(path) }.flatten.compact
+        end
+
+        def collect_files(path)
+          if File.directory?(path)
+            Dir.entries(path).map do |new_path|
+              next if [".", ".."].include?(new_path)
+              collect_files File.join(path, new_path)
```
GordonDiggs: nice

GordonDiggs:
LGTM

BlakeWilliams:
Thanks for the review @GordonDiggs!

PR #32, opened 2015-10-30
Don't detect comments as duplication

jpignata:
Scrubs innerComment nodes from the parse result to prevent comments from being considered duplication when analyzing JavaScript.


brynary:
Looks good

This PR actually reminds me that one thing people have had problems with for years is literal data structures getting flagged as duplication. e.g.:

```ruby
{
  foo: 1,
  bar: 2,
  # ...
}
```

We should probably exclude data structure nodes from being possible duplications too, so seems like something we want to be able to do in the general duplication processing code (not language specific). Nothing to do in this PR for that case, but wanted to make you aware of it.

PR #33, opened 2015-10-30
Add concurrency

brynary:
Adds thread-based concurrency to the parsing and `process_sexp` phase (but not the "report" phase at the end). Concurrency is implemented using a vanilla Ruby `Queue` and threads. The concurrency is configurable in the engine config (which I think we will want to add to the spec, and defaults to 2).

This uses the `Concurrent::Array`, `Concurrent::Hash` and `Concurrent::Map` classes from the `concurrent-ruby` gem for thread safe data structures in Flay (which is why I unpacked `flay.rb` from the Gem dependency to patch it.

We should consider submitting a concurrency patch upstream after we test this more. It will require some re-work but it's manageable.

_Note:_ The last step is switching the Ruby interpreter from MRI to JRuby in the Dockerfile (but this should produce wins even on MRI for non-Ruby languages due to the concurrent parsing in the e.g. Node.js processes we boot).

/c @jpignata

### DIFF
--- a/lib/cc/engine/analyzers/engine_config.rb
+++ b/lib/cc/engine/analyzers/engine_config.rb
@@ -14,6 +14,10 @@ module CC
           config.fetch("languages", {})
         end
 
+        def concurrency
+          config.fetch("concurrency", 2)
+        end
+
         def mass_threshold_for(language)
           threshold = fetch_language(language).fetch("mass_threshold", nil)
 
--- a/lib/cc/engine/analyzers/file_thread_pool.rb
+++ b/lib/cc/engine/analyzers/file_thread_pool.rb
@@ -1,0 +1,58 @@
+require "thread"
+
+module CC
+  module Engine
+    module Analyzers
+      class FileThreadPool
+        DEFAULT_CONCURRENCY = 2
+        MAX_CONCURRENCY = 2
+
+        def initialize(files, concurrency: DEFAULT_CONCURRENCY)
+          @files = files
+          @concurrency = concurrency
+        end
+
+        def run(&block)
+          queue = build_queue
+
+          @workers = thread_count.times.map do
+            Thread.new do
+              begin 
+                while item = queue.pop(true)
+                  yield item
+                end
+              rescue ThreadError
+              end
+            end
+          end
+        end
+
+        def join
+          workers.map(&:join)
+        end
+
+        private
+
+        attr_reader :files, :concurrency, :workers
+
+        def build_queue
+          Queue.new.tap do |queue|
+            files.each do |file|
+              queue.push(file)
+            end
+          end
+        end
+
+        def thread_count
+          if (1..MAX_CONCURRENCY) === concurrency
+            concurrency
+          elsif concurrency < 1
+            DEFAULT_CONCURRENCY
+          else
+            DEFAULT_CONCURRENCY
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/cc/engine/analyzers/reporter.rb
+++ b/lib/cc/engine/analyzers/reporter.rb
@@ -1,23 +1,36 @@
 require 'cc/engine/analyzers/violation'
+require 'cc/engine/analyzers/file_thread_pool'
+require 'thread'
 
 module CC
   module Engine
     module Analyzers
       class Reporter
         TIMEOUT = 10
-        def initialize(language_strategy, io)
+
+        def initialize(engine_config, language_strategy, io)
+          @engine_config = engine_config
           @language_strategy = language_strategy
           @io = io
         end
 
         def run
-          sexps = language_strategy.run
+          process_files
+          report
+        end
 
-          sexps.each do |sexp|
+        def process_files
+          pool = FileThreadPool.new(
+            language_strategy.files,
+            concurrency: engine_config.concurrency
+          )
+
+          pool.run do |file|
+            sexp = language_strategy.run(file)
             process_sexp(sexp)
           end
 
-          report
+          pool.join
         end
 
         def report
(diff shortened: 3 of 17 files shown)


brynary:
The CircleCI failure here looks unrelated.

lib/cc/engine/analyzers/engine_config.rb
```diff
@@ -14,6 +14,10 @@ def languages
           config.fetch("languages", {})
         end
 
+        def concurrency
```
BlakeWilliams: Is this something we want users to be able to configure?

lib/cc/engine/analyzers/reporter.rb
```diff
@@ -1,22 +1,45 @@
...
 
 module CC
   module Engine
     module Analyzers
       class Reporter
         TIMEOUT = 10
-        def initialize(language_strategy, io)
+
+        def initialize(config, language_strategy, io)
+          @config = config
```
BlakeWilliams: Instead of passing the entire config, thoughts on passing just the concurrency?

lib/cc/engine/analyzers/reporter.rb
```diff
@@ -1,22 +1,45 @@
...
-          sexps = language_strategy.run
+          queue = Queue.new
 
-          sexps.each do |sexp|
-            process_sexp(sexp)
+          language_strategy.each_file do |file|
+            queue.push(file)
           end
 
+          workers = @config.concurrency.times.map do
```
BlakeWilliams: Can this be extracted into another method to make `run` more readable?

lib/cc/engine/analyzers/reporter.rb
```diff
@@ -1,22 +1,45 @@
...
+            Thread.new do
+              begin
+                while file = queue.pop(true)
+                  begin
```
BlakeWilliams: Could the begin/rescue/end for processing files also be extracted into another method?

lib/cc/engine/analyzers/engine_config.rb
GordonDiggs: Yes definitely, though we should probably consider having some max in place

lib/cc/engine/analyzers/reporter.rb
BlakeWilliams: Actually, I think you could keep `run` on analyzer base and make it accept a single file and pass it in which means you can keep the file processing logic in the analyzers and just do:

```
begin
  sexp = language_strategy.run(file)
  process_sexp(sexp) if sexp
end
```

lib/cc/engine/analyzers/reporter.rb
```diff
@@ -1,22 +1,45 @@
...
+                    sexp = language_strategy.process_file(file)
+                  rescue => ex
+                    $stderr.puts "Skipping file #{file} due to exception"
+                    $stderr.puts "(#{ex.class}) #{ex.message} #{ex.backtrace.join("\n")}"
+                  end
+
+                  process_sexp(sexp) if sexp
+                end
+              rescue ThreadError
```
BlakeWilliams: Maybe log to stderr?

BlakeWilliams:
Instead of pulling in flay directly, what do you think about forking flay and making changes in that repo?

lib/cc/engine/analyzers/engine_config.rb
BlakeWilliams: Yeah, that was my concern. Don't want someone going overboard with concurrency.

jpignata: I'd cap it at 2 for now. Later, we'll probably make this externally configurable by a build runner so that we can set the max to other values.

BlakeWilliams: :+1:

BlakeWilliams:
Alright, added some changes:

* Refactored concurrency into `FileThreadPool`
* Extracted flay changes to https://github.com/codeclimate/flay

BlakeWilliams:
cc @codeclimate/review

lib/cc/engine/analyzers/file_thread_pool.rb
```diff
@@ -0,0 +1,56 @@
...
+        end
+
+        def run(&block)
+          queue = build_queue
+
+          @workers = thread_count.times.map do
+            Thread.new do
+              begin 
+                yield queue.pop(true)
+              rescue ThreadError
```
MattMSumner: Since we're rescuing here, how would you feel about logging that this happened?

BlakeWilliams: Agreed, I'll print something to stderr.

lib/cc/engine/analyzers/file_thread_pool.rb
```diff
@@ -0,0 +1,56 @@
...
+        def build_queue
+          Queue.new.tap do |queue|
+            files.each do |file|
+              queue.push(file)
+            end
+          end
+        end
+
+        def thread_count
+          if concurrency > MAX_CONCURRENCY
```
MattMSumner: how would you feel about using a range comparison here? http://ruby-doc.org/core-2.2.0/Range.html#method-i-3D-3D-3D

```rb
if (1..MAX_CONCURRENCY) === concurrency
  concurrency
else
  DEFAULT_CONCURRENCY
end
```

MattMSumner:
Few comments otherwise this LGTM.

lib/cc/engine/analyzers/file_thread_pool.rb
BlakeWilliams: Love it!

lib/cc/engine/analyzers/file_thread_pool.rb
brynary: This rescue is actually just for when queue.pop runs out of files to pop, so I don't think it should log to stderr.

PR #34, opened 2015-11-02
Add fingerprint to violation json

BlakeWilliams:
This uses the first filename and the mass of the s-expression as the
fingerprint for duplications.

This also changes the tests to test properties instead of a comparing against a fixture like string.

### DIFF
--- a/lib/cc/engine/analyzers/violation.rb
+++ b/lib/cc/engine/analyzers/violation.rb
@@ -1,3 +1,5 @@
+require "digest"
+
 module CC
   module Engine
     module Analyzers
@@ -19,7 +21,8 @@ module CC
             "location": format_location,
             "remediation_points": calculate_points,
             "other_locations": format_other_locations,
-            "content": content_body
+            "content": content_body,
+            "fingerprint": fingerprint
           }
         end
 
@@ -28,7 +31,11 @@ module CC
         attr_reader :base_points, :hashes
 
         def current_sexp
-          @location ||= hashes.first
+          @location ||= sorted_hashes.first
+        end
+
+        def sorted_hashes
+          @_sorted_hashes ||= hashes.sort_by(&:file)
         end
 
         def other_sexps
@@ -86,6 +93,16 @@ module CC
           File.expand_path(
             File.join(File.dirname(__FILE__), relative_path)
           )
+        end
+
+        def fingerprint
+          digest = Digest::MD5.new
+          digest << current_sexp.file
+          digest << "-"
+          digest << current_sexp.mass.to_s
+          digest << "-"
+          digest << occurrences.to_s
+          digest.to_s
         end
 
         def description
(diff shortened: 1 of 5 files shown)


BlakeWilliams:
cc @codeclimate/review

lib/cc/engine/analyzers/violation.rb
```diff
@@ -19,7 +19,8 @@ def format
             "location": format_location,
             "remediation_points": calculate_points,
             "other_locations": format_other_locations,
-            "content": content_body
+            "content": content_body,
+            "fingerprint": "#{current_sexp.file}-#{current_sexp.mass}"
```
GordonDiggs: Fingerprints are generally hashes. What do you think of having a `fingerprint` method that hashes this string?

BlakeWilliams:
@GordonDiggs One more look?

wfleming:
This is a good direction, @BlakeWilliams. Are you going to do the other changes separately to take into account number of files the issue appears in, and doing some sorting to so we won't see false "fixed/new" alerts anymore?

lib/cc/engine/analyzers/violation.rb
```diff
@@ -88,6 +95,12 @@ def read_up_path
           )
         end
 
+        def fingerprint
+          digest = Digest::MD5.new
+          digest << "#{current_sexp.file}-#{current_sexp.mass}"
```
GordonDiggs: I think these should be separate additions. There's no benefit in constructing the string and then calling `<<`

BlakeWilliams:
We do sort the files now, but I didn't add the number of issues. I'll add that in.

wfleming:
@BlakeWilliams Great, thanks! :+1:

BlakeWilliams:
Feedback addressed.

lib/cc/engine/analyzers/violation.rb
```diff
@@ -88,6 +95,16 @@ def read_up_path
           )
         end
 
+        def fingerprint
+          digest = Digest::MD5.new
+          digest << current_sexp.file
+          digest << "-"
```
GordonDiggs: These aren't needed because the whole thing is being hashed and the digest library itself handles the chunks

BlakeWilliams: Gotcha, I took that part since that's what the link you gave me did as well. I'll remove em'.

GordonDiggs: Oh weird - missed that. I think it's fine then! Whatever the old one does we should match

GordonDiggs:
LGTM

lib/cc/engine/analyzers/violation.rb
BlakeWilliams: Sure, I'll keep it then. :smiley:

PR #35, opened 2015-11-03
Reduce node generated s-expression mass

BlakeWilliams:
Instead of creating an s-expression for each `key` => `value` in an
s-expression node we insert only `value` directly as a symbol which
reduces the s-expression count which reduces the mass.

We do this because the s-expression tree that is currently being
generated feels too heavy and can make trivial code mass higher than the
threshold when it should be below the threshold.

eg:

```
const hello = function(name) {
  console.log(`${name}!`);
};

hello("world");
```

The above code currently has a mass of 54. With this change
it's now has a mass of 41.

### DIFF
--- a/lib/cc/engine/analyzers/node.rb
+++ b/lib/cc/engine/analyzers/node.rb
@@ -45,7 +45,7 @@ module CC
             elsif  value.is_a?(Hash)
               create_sexp(key.to_sym, self.class.new(value, @file, @line).format)
             else
-              create_sexp(key.to_sym, value)
+              value.to_s.to_sym
             end
           end
         end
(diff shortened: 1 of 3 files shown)


pbrisbin:
Code seems fine, but I'm not sure I understand the reasoning behind the change. The commit message is detailed, but seems to be focused on the What not the Why. Can you explain the motivation a bit, and ammend it into the commit message too?

BlakeWilliams:
Sure thing.

jpignata:
LGTM

pbrisbin:
I suspect this causes something to be flagged as duplication that shoudn't be. Is there an example of code that we don't want to be considered duplicated that is before this fix?

BlakeWilliams:
> I suspect this causes something to be flagged as duplication that shoudn't be.

Due to mass threshold? I think it would mark more violations as `Similar code` instead of always being `Duplicated code`.

I don't have any examples at the moment, I'll try to put something together and share the results.

pbrisbin:
Hmm... so I guess that's what I was getting at with wanting to know the Why.

Not a merge-blocker, but I still don't feel like I fully understand what the "too heavy" mass was causing that was undesirable and needed fixing -- it's possible future commit-message readers won't understand either.

BlakeWilliams:
It's largely a "does it feel right" thing, so it's a bit hard to explain exactly why since it's not 100% quantifiable.

PR #36, opened 2015-11-04
Switch from MRI to JRuby

BlakeWilliams:
To take full advantage of concurrent-ruby for faster analyses this
replaces MRI Ruby with JRuby since JRuby can take advantage of threads
without being blocked by the GIL.

* Replace posix-spawn with `IO.popen`
* Replace alpine with `jruby:9.0-jdk` Docker image

Here's a comparison of JRuby vs MRI on a medium sized JavaScript
project using the same concurrency value and same config.

JRuby:

```
codeclimate analyze --dev  0.29s user 0.22s system 0% cpu 2:26.03 total
```

MRI:

```
codeclimate analyze --dev  0.35s user 0.29s system 0% cpu 3:54.92 total
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,4 @@
-
-FROM alpine:edge
+FROM jruby:9.0.3-jdk
 
 WORKDIR /usr/src/app/
 
@@ -9,19 +8,19 @@ COPY Gemfile.lock /usr/src/app/
 COPY vendor/php-parser/composer.json /usr/src/app/vendor/php-parser/
 COPY vendor/php-parser/composer.lock /usr/src/app/vendor/php-parser/
 
-RUN apk --update add openssh git python nodejs php-cli php-json php-phar php-openssl php-xml curl\
-    ruby ruby-io-console ruby-dev build-base && \
-    gem install bundler --no-ri --no-rdoc && \
+RUN curl --silent --location https://deb.nodesource.com/setup_5.x | bash -
+RUN apt-get install -y nodejs python openssh-client php5-cli php5-json
+RUN gem install bundler --no-ri --no-rdoc && \
     bundle install -j 4 && \
-    apk del build-base && rm -fr /usr/share/ri && \
     curl -sS https://getcomposer.org/installer | php
 
 RUN mv composer.phar /usr/local/bin/composer
 RUN cd /usr/src/app/vendor/php-parser/ && composer install --prefer-source --no-interaction
 
-RUN adduser -u 9000 -D app
+RUN adduser app -u 9000
 
 COPY . /usr/src/app
+run chown -R app .
 RUN npm install
 
 USER app
--- a/lib/cc/engine/analyzers/python/parser.rb
+++ b/lib/cc/engine/analyzers/python/parser.rb
@@ -1,4 +1,3 @@
-require 'posix/spawn'
 require 'timeout'
 require 'json'
 
@@ -42,10 +41,14 @@ module CC
 
           def run(input, timeout = DEFAULT_TIMEOUT)
             Timeout.timeout(timeout) do
-              child = ::POSIX::Spawn::Child.new(command, input: input, timeout: timeout)
+              IO.popen command, "r+" do |io|
+                io.puts input
+                io.close_write
 
-              if child.status.success?
-                yield child.out if block_given?
+                output = io.gets
+                io.close
+
+                yield output if $?.to_i == 0
               end
             end
           end
--- a/Gemfile
+++ b/Gemfile
@@ -4,7 +4,6 @@ gem 'flay', git: 'https://github.com/codeclimate/flay.git'
 gem 'concurrent-ruby', "~> 1.0.0.pre4"
 gem 'ruby_parser'
 gem 'pry'
-gem 'posix-spawn'
 gem 'sexp_processor'
 gem 'json'
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,6 +1,6 @@
 GIT
   remote: https://github.com/codeclimate/flay.git
-  revision: e6e472e064dc459277f8bd57167789f8fed77f56
+  revision: ac8fd46ba30e9636e8bdc625cf5f849615d505b6
   specs:
     flay (2.6.1)
 
@@ -8,11 +8,10 @@ GEM
   remote: https://rubygems.org/
   specs:
     coderay (1.1.0)
-    concurrent-ruby (1.0.0.pre4)
+    concurrent-ruby (1.0.0.pre5)
     diff-lcs (1.2.5)
     json (1.8.3)
     method_source (0.8.2)
-    posix-spawn (0.3.11)
     pry (0.10.3)
       coderay (~> 1.1.0)
       method_source (~> 0.8.1)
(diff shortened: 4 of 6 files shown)


BlakeWilliams:
cc @codeclimate/review

Dockerfile
```diff
@@ -9,19 +8,21 @@ COPY Gemfile.lock /usr/src/app/
...
-RUN apk --update add openssh git python nodejs php-cli php-json php-phar php-openssl php-xml curl\
-    ruby ruby-io-console ruby-dev build-base && \
-    gem install bundler --no-ri --no-rdoc && \
+RUN curl --silent --location https://deb.nodesource.com/setup_5.x | bash -
+RUN apt-get install -y nodejs python openssh-client php5-cli php5-json
+RUN gem install bundler --no-ri --no-rdoc && \
     bundle install -j 4 && \
-    apk del build-base && rm -fr /usr/share/ri && \
     curl -sS https://getcomposer.org/installer | php
+# RUN apk --update add openssh git python nodejs php-cli php-json php-phar php-openssl php-xml curl\
```
jpignata: Is this work in progress or accidentally left commented out?

Dockerfile
```diff
@@ -1,5 +1,4 @@
-
-FROM alpine:edge
+FROM jruby:9.0.3-jdk
```
GordonDiggs: What's the size difference between these 2 images?

brynary:
How big is that repo? The repo we are targeting the optimization at takes over 15 minutes to analyze (it's JSX)

jpignata:
relax/relax took 8 minutes previously – that might be a good one to benchmark.

Dockerfile
BlakeWilliams: Oops, missed it.

lib/cc/engine/analyzers/python/parser.rb
```diff
@@ -42,10 +41,14 @@ def initialize(command, delegate)
 
           def run(input, timeout = DEFAULT_TIMEOUT)
             Timeout.timeout(timeout) do
-              child = ::POSIX::Spawn::Child.new(command, input: input, timeout: timeout)
+              IO.popen command, 'r+' do |io|
```
jpignata: Styleguide :lipstick: single quotes -> double quotes (though, they seem to be used inconsistently throughout this repo).

BlakeWilliams: Good catch, I think it started with single and I've been ~~failing~~ attempting to move it over to double.

Dockerfile
BlakeWilliams: JRuby come in pretty big at 712mb.

BlakeWilliams:
I'm running it on relax/relax now. I'll report back with the results.

BlakeWilliams:
JRuby:

```
Analysis complete! Found 393 issues.
codeclimate analyze --dev  0.39s user 0.34s system 0% cpu 4:24.52 total
```

MRI:

```
Analysis complete! Found 401 issues.
codeclimate analyze --dev  0.51s user 0.49s system 0% cpu 7:12.91 total
```

The interesting part is that JRuby has less issues than MRI. Could this be a cause of using [`Map`](http://ruby-concurrency.github.io/concurrent-ruby/Concurrent/Map.html) instead of `Hash` in [our flay fork](https://github.com/codeclimate/flay/blob/master/lib/flay.rb#L217)? Less likely, but maybe it could be a result of using `IO.popen` instead of posix-spawn?

BlakeWilliams:
Here's the config I used:

```
engines:
  duplication:
    enabled: true
    config:
      languages:
        javascript:
          paths:
            - "lib/**/*.js"
            - "lib/**/*.jsx"
ratings:
  paths:
  - "**.js"
```

BlakeWilliams:
After doing a bit of testing, I think the difference in results comes from the `Array#hash` implementation differing between MRI/JRuby.

BlakeWilliams:
Okay, found a crazy workaround where we now monkey patch `Array#hash` in JRuby to get similar results in as you would in MRI. I'm not sure what version of murmurhash MRI uses, but we pull in [murmurhash3](https://github.com/funny-falcon/murmurhash3-ruby) and run it against `Array#to_s`.

Here's the results from JRuby + the monkey patch:

```
Analysis complete! Found 401 issues.
codeclimate analyze --dev  0.38s user 0.34s system 0% cpu 4:35.15 total
```

fhwang:
Is there a full explanation somewhere of why this workaround works? Or can you write one out here? ;) It's a little surprising to me because equality should be the same across implementations regardless of how hashcodes are generated.

I try to be cautious about changing low-level behavior like this. It's nice that this monkeypatch would be contained within the limited scope of an engine, but I also wonder if we couldn't do something less drastic instead, such as further modifying our fork of `flay`.

BlakeWilliams:
The commit message + that commit is as much as I know.

tl;dr JRuby does it differently so we use murmurhash like MRI does.

I was thinking about changing our Flay fork but I wasn't sure if it made sense because this seems like an implementation bug and not really an improvement or fix. As for limited scope, it's only in the scope of this engine when running JRuby. It's not patched for MRI since it has the correct behavior.

BlakeWilliams:
This looks like the implementation of JRuby's `Array#hash` https://github.com/jruby/jruby/blob/master/core/src/main/java/org/jruby/RubyArray.java#L667-L685

And here's MRI's: http://rxr.whitequark.org/mri/source/array.c?v=2.1.2#3796

fhwang:
K, from closer reading of that commit I think we may have simply found a bug in Flay.

If I remember correctly (I'm not having luck finding this documented), `Object#hashcode` in Ruby is guaranteed to be deterministic, but it is not guaranteed to be unique. It's effectively unique most of the time, since hashcodes are large, but it is possible for two unequal objects to return the same hashcode. (Which makes sense when you think about it, as hashcodes are an integer of fixed size whereas the space of all possible objects in Ruby is basically infinite.)

So if a library like `Flay` is using hashcodes to determine uniqueness, that's probably a bug. Any Ruby implementation should be able to use any sufficient hashcode algorithm as long as it satisfies the requirement of being deterministic.

To take off my pedantic hat, the main reason I'm trying to get clarity on this issue is because it would be nice to avoid such low-level tinkering if we can. You were able to find the fact that Flay depends on hashcodes -- does that mean you could also patch Flay to depend on something correct like straight-up array equality?

FWIW I love a good monkeypatch, just nothing this low-level ;)

BlakeWilliams:
Is `Array#hash` using `Object#hashcode` under the hood? From what I've seen it's using murmurhash generate the hash. I don't think that's an unreasonable thing for flay to do if you look at how it works under the hood.

I think calling this a flay bug is a bit of a stretch since this stems from JRuby not using the same method to generate a hash from an array as MRI does.

I agree that we should avoid low-level tinkering, and I also think monkey patches are awful but this is only here because it's a workaround for a broken implementation of `Array#hash`.

BlakeWilliams:
We removed the monkey patch in favor of updating our flay fork to md5 the s-expressions instead of using murmur or `Array#hash`.

Any more feedback or is this good to go?

fhwang:
LGTM

BlakeWilliams:
Thanks!

PR #37, opened 2015-11-09
Don't report same line for multiple duplications

BlakeWilliams:
This uses Ruby's [`set`](http://ruby-doc.org/stdlib-2.2.3/libdoc/set/rdoc/Set.html) class
to keep track of which lines in every file is reported to prevent
reporting the same line multiple times.

This happens most frequently with JavaScript since the Babel generated
ast is verbose and can expand a single line into a large s-expression
tree which can have several duplications.

### DIFF
--- a/lib/cc/engine/analyzers/violation.rb
+++ b/lib/cc/engine/analyzers/violation.rb
@@ -26,6 +26,10 @@ module CC
           }
         end
 
+        def report_name
+          "#{current_sexp.file}-#{current_sexp.line}"
+        end
+
         private
 
         attr_reader :base_points, :hashes
--- a/spec/cc/engine/analyzers/javascript/main_spec.rb
+++ b/spec/cc/engine/analyzers/javascript/main_spec.rb
@@ -58,6 +58,16 @@ RSpec.describe CC::Engine::Analyzers::Javascript::Main do
       expect(run_engine(engine_conf)).to be_empty
   end
 
+  it "does not report the same line for multiple issues" do
+    create_source_file("dup.jsx", <<-EOJSX)
+          <a className='button button-primary full' href='#' onClick={this.onSubmit.bind(this)}>Login</a>
+    EOJSX
+
+    result = run_engine(engine_conf).strip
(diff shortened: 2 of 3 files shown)


lib/cc/engine/analyzers/violation.rb
```diff
@@ -26,6 +26,10 @@ def format
           }
         end
 
+        def report_name
+          "#{current_sexp.file}-#{current_sexp.line}"
```
GordonDiggs: Should this include end line as well?

BlakeWilliams: I was thinking about that myself, and I'm not sure honestly. I don't think there would be any (or very many at the least) instances where the same start line has multiple end-lines.

I'd love to hear what others think about that though.

GordonDiggs: Cool - You know more about this so I defer to you on that!

GordonDiggs:
LGTM

spec/cc/engine/analyzers/javascript/main_spec.rb
```diff
@@ -58,6 +58,16 @@
...
   end
 
+  it "does not report the same line for multiple issues" do
+    create_source_file("dup.jsx", <<-EOJSX)
+          <a className='button button-primary full' href='#' onClick={this.onSubmit.bind(this)}>Login</a>
+    EOJSX
+
+    result = run_engine(engine_conf).strip
+    issues = result.split("\0")
+    expect(issues.length).to eq 1
```
GordonDiggs: @BlakeWilliams can you explain why this is 1? Is it because we set the mass to 1 for the tests? I wouldn't expect a line to be reported unless there was another instance of it

BlakeWilliams: Yep, it's because we pass in an extremely low mass threshold so this will always be higher.

GordonDiggs: Ok thanks!

PR #38, opened 2015-11-10
Don't report same line in other_locations

BlakeWilliams:
This follows up on the PR for not reporting multiple duplications for
the same line and applied the same logic to `other_locations`.


GordonDiggs:
LGTM

PR #39, opened 2015-11-10
Don't show snippet in other locations

GordonDiggs:
This reverts the previous attempt to exclude the current snippet from `other_locations`.
It's not always the first in `hashes`, but `current_sexp` uses the first
of the sorted array, so dropping that one is safe.

Covered by tests here:
https://github.com/codeclimate/codeclimate-duplication/blob/master/spec/cc/engine/analyzers/javascript/main_spec.rb#L41-42

@codeclimate/review

### DIFF
--- a/lib/cc/engine/analyzers/violation.rb
+++ b/lib/cc/engine/analyzers/violation.rb
@@ -6,11 +6,10 @@ module CC
       class Violation
         attr_reader :issue
 
-        def initialize(base_points, issue, hashes, reports)
+        def initialize(base_points, issue, hashes)
           @base_points = base_points
           @issue = issue
           @hashes = hashes
-          @reports = reports
         end
 
         def format
@@ -33,7 +32,7 @@ module CC
 
         private
 
-        attr_reader :base_points, :hashes, :reports
+        attr_reader :base_points, :hashes
 
         def current_sexp
           @location ||= sorted_hashes.first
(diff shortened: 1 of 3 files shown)


lib/cc/engine/analyzers/violation.rb
```diff
@@ -33,7 +32,7 @@ def report_name
 
         private
 
-        attr_reader :base_points, :hashes, :reports
+        attr_reader :base_points, :hashes
 
         def current_sexp
           @location ||= sorted_hashes.first
```
GordonDiggs: This is where it uses sorted_hashes

GordonDiggs:
screen shot 2015-11-10 at 11 29 37 am

Sample result

BlakeWilliams:
Out of curiosity, can you try this with two duplicated lines of JSX?

```
```

I think since that has multiple duplications in the same line `other_locations` might get triggered, but I'm not sure.

BlakeWilliams:
LGTM

jpignata:
:+1:

GordonDiggs:
@BlakeWilliams pasting that line twice worked correctly with these changes. Though, apparently 2 lines like that without a wrapping div is invalid JSX

BlakeWilliams:
Cool, good to know. And yeah, JSX requires one top-level div.

PR #40, opened 2015-11-11
Use babylon parser directly to disable strict mode

dblandin:
We shouldn't assume that all JavaScript analyzed by this engine will adhere to strict ES6 rules. Instead, use the parser as loosely as possible when analyzing for duplication.

```
/usr/src/app/node_modules/babel/node_modules/babel-core/lib/transformation/file/index.js:671
      throw err;
      ^

SyntaxError: unknown: Argument name clash in strict mode (54:39)
  52 |       chart.yAxis.tickFormat(function() { return ""; });
  53 |
> 54 |       chart.tooltipContent(function(_, _, _, graph) {
     |                                        ^
  55 |         var date = d3.time.format('%b-%d')(new Date(graph.point.x));
  56 |         return "<strong><span style=\"padding:5px;\">" + date + "</span></strong>";
  57 |       });
    at Parser.pp.raise (/usr/src/app/node_modules/babel/node_modules/babel-core/node_modules/babylon/lib/parser/location.js:24:13)
    at Parser.pp.checkLVal (/usr/src/app/node_modules/babel/node_modules/babel-core/node_modules/babylon/lib/parser/lval.js:181:16)
    at Parser.pp.parseFunctionBody (/usr/src/app/node_modules/babel/node_modules/babel-core/node_modules/babylon/lib/parser/expression.js:766:12)
    at Parser.parseFunctionBody (/usr/src/app/node_modules/babel/node_modules/babel-core/node_modules/babylon/lib/plugins/flow.js:609:20)
    at Parser.pp.parseFunction (/usr/src/app/node_modules/babel/node_modules/babel-core/node_modules/babylon/lib/parser/statement.js:524:8)
    at Parser.pp.parseExprAtom (/usr/src/app/node_modules/babel/node_modules/babel-core/node_modules/babylon/lib/parser/expression.js:397:19)
    at Parser.parseExprAtom (/usr/src/app/node_modules/babel/node_modules/babel-core/node_modules/babylon/lib/plugins/jsx/index.js:412:22)
    at Parser.pp.parseExprSubscripts (/usr/src/app/node_modules/babel/node_modules/babel-core/node_modules/babylon/lib/parser/expression.js:236:19)
    at Parser.pp.parseMaybeUnary (/usr/src/app/node_modules/babel/node_modules/babel-core/node_modules/babylon/lib/parser/expression.js:217:19)
[...]

PR #41, opened 2015-11-11
Change directory with a block while running the reporter

dblandin:
I experienced some weirdness recently with the error described in #40 showing up multiple times in the container logs. Using the block form of chdir seems to have removed that repetition, with just one expected occurrence showing up. I wonder if there's some strangess around changing directories and multiple jruby threads that I still don't understand, but using the block form seems to be a safe move.

Thoughts?

### DIFF
--- a/lib/cc/engine/duplication.rb
+++ b/lib/cc/engine/duplication.rb
@@ -22,22 +22,24 @@ module CC
       }.freeze
 
       def initialize(directory:, engine_config:, io:)
-        Dir.chdir(directory)
+        @directory = directory
         @engine_config = CC::Engine::Analyzers::EngineConfig.new(engine_config || {})
         @io = io
       end
 
       def run
-        languages_to_analyze.each do |language|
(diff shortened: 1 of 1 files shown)


lib/cc/engine/duplication.rb
```diff
@@ -31,13 +31,16 @@ def run
         languages_to_analyze.each do |language|
           engine = LANGUAGES[language].new(engine_config: engine_config)
           reporter = CC::Engine::Analyzers::Reporter.new(engine_config, engine, io)
-          reporter.run
+
+          Dir.chdir(directory) do
```
GordonDiggs: Would it work to wrap the entire Duplication#run body in the `chdir`? That is closer to what we do in other repos and also closer to the original intention of doing it in the constructor

dblandin: Sure, I'll make that change. :+1:

GordonDiggs:
:+1:

PR #42, opened 2015-11-13
Set JVM heap size options

jpignata:
We're seeing the duplication engine being killed for memory consumption
in production while analyzing medium-sized repositories. This change
sets an initial and maximum heap space of 384MB. This will keep us
under the memory limit while leaving additional working memory for
other processes.

--

This is kind of a shot in the dark. We're seeing this behavior only in production and not in the CLI. Thoughts, @codeclimate/review?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,4 +25,6 @@ RUN npm install
 
 USER app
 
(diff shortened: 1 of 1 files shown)


GordonDiggs:
Worth a shot!

PR #44, opened 2015-11-18
improve derivation of lines from Sexp

wfleming:
This started as a hunt into why an issue was reported for just a single
line saying `end`, with the only other instance of the duplication being
the same line: the source code in the spec is a reduced instance of the
triggering code.

The poor behavior results from how RubyParser/Sexp handle certain
structures: the "line" of the `rescue..end` expression is parsed as the
location of the `end`. This seems wrong, but I can't figure out a quick
fix within those libraries.

The way child sexprs get lines seems generally sane: at the very small level,
lines seem pretty reliable. Strange behavior emerges at the higher
level. The way we were calculating the end line of an sexpr already
implicitly acknowledged this by walking the whole tree to find the
highest line.

This change adapts that previous algorithm to apply the same principle
to finding both the start & end lines: the start line is basically the
smallest line of the sexpr or any of its children, and the end line is
the maximum.

:eyes: @codeclimate/review

### DIFF
--- a/lib/cc/engine/analyzers/sexp_lines.rb
+++ b/lib/cc/engine/analyzers/sexp_lines.rb
@@ -1,0 +1,28 @@
+module CC
+  module Engine
+    module Analyzers
+      class SexpLines
+        attr_reader :begin_line, :end_line
+
+        def initialize(root_sexp)
+          @root_sexp = root_sexp
+          calculate
+        end
+
+        private
+
+        attr_reader :root_sexp
+
+        def calculate
+          @begin_line = root_sexp.line
+          @end_line = root_sexp.end_line || root_sexp.line
+
+          root_sexp.deep_each do |sexp|
+            @begin_line = [@begin_line, sexp.line].min
+            @end_line = [@end_line, sexp.end_line || sexp.line].max
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/cc/engine/analyzers/violation.rb
+++ b/lib/cc/engine/analyzers/violation.rb
@@ -69,23 +69,14 @@ module CC
         end
 
         def format_sexp(sexp)
+          lines = SexpLines.new(sexp)
           {
             "path": sexp.file.gsub(%r(^./), ""),
             "lines": {
-              "begin": sexp.line,
-              "end": sexp.end_line || sexp_max_line(sexp, sexp.line)
-            }
+              "begin": lines.begin_line,
+              "end": lines.end_line,
+            },
           }
-        end
-
-        def sexp_max_line(sexp_tree, default)
-          max = default
-
-          sexp_tree.deep_each do |sexp|
-            max = sexp.line if sexp.line > max
-          end
-
-          max
         end
 
         def content_body
--- a/spec/cc/engine/analyzers/sexp_lines_spec.rb
+++ b/spec/cc/engine/analyzers/sexp_lines_spec.rb
@@ -1,0 +1,43 @@
+require "spec_helper"
+require "cc/engine/duplication"
+
+module CC::Engine::Analyzers
+  RSpec.describe SexpLines do
+    describe "violation location" do
+      it "gets appropriate locations for rescue blocks" do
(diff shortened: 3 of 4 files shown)


spec/cc/engine/violation_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+        flay.process_sexp(sexp)
+
+        report = flay.analyze[0]
+        violation = Violation.new(10, report, flay.hashes[report.structural_hash])
+
+        location = violation.send(:format_location)
+        other_location = violation.send(:format_other_locations)[0]
+
+        expect(location).not_to eq other_location
+        expect(location[:lines]).to eq({:begin => 3, :end => 7})
```
wfleming: The overlap between these two nodes is a little weird, but makes sense from a parser sense: syntactically, a `rescue` expr doesn't end until the `end`: the next `rescue` is a distinct node, but the two do overlap in terms of lines.

I've opted not to do anything about that right now because 1) it's not a super common case for code duplication and 2) I don't have a good answer for a generalizable automatic remediation. We could just drop the "overlapping" lines from the bigger node (i.e. the above location would become `{:begin => 3, :end => 4}`, but I'm not confident there aren't other cases of node overlap that would result in even more nonsensical display.

spec/cc/engine/violation_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+
+    describe "violation location" do
+      it "gets appropriate locations for rescue blocks" do
+        sexp = RubyParser.new.process(source, "file.rb")
+        flay.process_sexp(sexp)
+
+        report = flay.analyze[0]
+        violation = Violation.new(10, report, flay.hashes[report.structural_hash])
+
+        location = violation.send(:format_location)
```
wfleming: I'm calling private methods here in the interest of only testing locations, which really is all I care about here. I can shift to calling `format` and inspecting those results if people prefer.

lib/cc/engine/analyzers/violation.rb
```diff
@@ -69,23 +69,25 @@ def format_other_locations
...
-              "end": sexp.end_line || sexp_max_line(sexp, sexp.line)
+              "begin": lines[0],
+              "end": lines[1]
             }
           }
         end
 
-        def sexp_max_line(sexp_tree, default)
-          max = default
+        def sexp_lines(sexp_tree)
```
pbrisbin: This method is getting a little fiddly -- what do you think about a small class? The class could then have methods for the begin/end so you don't need the array return value either.

wfleming:
@pbrisbin refactored

lib/cc/engine/analyzers/sexp_lines.rb
```diff
@@ -0,0 +1,32 @@
...
+      class SexpLines
+        attr_reader :begin_line, :end_line
+
+        def initialize(root_sexp)
+          @root_sexp = root_sexp
+          calculate
+        end
+
+        def to_h
+          {"begin": begin_line, "end": end_line}
```
pbrisbin: What about leaving the hashifiying to `#format` in the other class? Feels like a "formatting" responsibility.

Then you can feel better testing the public `begin_line`/`end-line` methods that indeed need to be exposed.

wfleming: I did this just because it made `Violation#format_sexp` prettier. Can back out.

spec/cc/engine/analyzers/sexp_lines_spec.rb
```diff
@@ -0,0 +1,46 @@
+require "spec_helper"
+require "cc/engine/duplication"
+
+module CC::Engine::Analyzers
+  RSpec.describe SexpLines do
+    let(:source) do
```
pbrisbin: WDYT about inlining this into the example where it's used? It's only used there, and it's important to the spec, so I think it should be co-located.

pbrisbin:
LGTM

PR #45, opened 2015-11-18
Put Reported Issue's Mass in Read Up Contents

wfleming:
One aspect of the duplication engine that is different from other engines
and that can be frustrating is that the main dial you can turn to adjust
the results, mass threshold, is a completely invisible value once the issue
has been reported to CC (either .com or CLI). As a user, this makes the
engine difficult to configure because even though the README documents how
to set the mass threshold, the fact that you can't ever see what mass
issues have makes the feedback loop between getting results and deciding
what your settings should be basically impossible.

This is an attempt to remediate this problem by describing generally what
mass is, and what its specific value for an issue is, within the read up
contents. This makes the value accessible without unduly cluttering the
description with that value.

Also some other minor cleanup of copy & the test suite I ran into while I was doing this.

:eyes: @codeclimate/review


lib/cc/engine/analyzers/violation_read_up.rb
```diff
@@ -0,0 +1,28 @@
...
+        end
+
+        def contents
+          ERB.new(File.read(read_up_path)).result(binding)
+        end
+
+        private
+
+        def read_up_path
+          relative_path = "../../../../config/contents/duplicated_code.md.erb"
```
BlakeWilliams: Thoughts on making this a constant?

lib/cc/engine/analyzers/violation_read_up.rb
```diff
@@ -0,0 +1,28 @@
+require "erb"
+
+module CC
+  module Engine
+    module Analyzers
+      class ViolationReadUp
+        attr_reader :issue
```
BlakeWilliams: Thoughts on making this `attr_reader` private?

BlakeWilliams:
I like the extracted class! LGTM

PR #46, opened 2015-11-18
DRY analyzer specs

wfleming:
These pieces were repeated pretty much verbatim between the different
analyzer specs. That seems especially silly since this is, after all, an
engine for pointing out code duplication.

:eyes: @codeclimate/review

### DIFF
--- a/spec/cc/engine/analyzers/javascript/main_spec.rb
+++ b/spec/cc/engine/analyzers/javascript/main_spec.rb
@@ -5,16 +5,8 @@ require 'cc/engine/analyzers/file_list'
 require 'flay'
 require 'tmpdir'
 
-RSpec.describe CC::Engine::Analyzers::Javascript::Main do
-  around do |example|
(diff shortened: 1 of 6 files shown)


brynary:
This looks good, but did we consider moving these to use the new `codeclimate test` harness?

spec/support/analyzer_spec.rb
```diff
@@ -0,0 +1,28 @@
+module AnalyzerSpec
+  def self.included(base)
```
pbrisbin: Would this be better added via `RSpec.configure`? Maybe on a `in_tmpdir: true` tag or something?

wfleming:
@brynary That's a good idea, but not something I want to do just now: these specs test some other things like readup contents that I don't believe the test harness handles? It would be good to use the test harness & break the other things into different specs, but it's a bigger refactor than I want to get into right now.

pbrisbin:
LGTM

PR #53, opened 2015-12-17
remove Java heap settings

wfleming:
These seem to be causing non-deterministic analysis results from some
behavior in JRuby/the JVM. While we hunt this down, it would be
preferable to not set heap limits & let the engine hard-crash by hitting
container memory limits.

cc @codeclimate/review

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,6 +25,4 @@ RUN npm install
 
 USER app
 
-ENV JAVA_OPTS="-Xms128m -Xmx128m"
-
 CMD ["/usr/src/app/bin/duplication"]


pbrisbin:
Makes sense to me. :shipit:

brynary:
Are you sure you want to remove both the min and the max? It sounds like your description is all about the max. Allocating a large enough heap at the beginning is generally faster, as it saves as-you-go heap expansions

jpignata:
That may be true, but I doubt 128MB is the right value for this. I think
I'd say we remove both for now as Will has done, and wait until our
investigation is complete to set these values more precisely.

brynary:
Makes sense

PR #54, opened 2015-12-17
Add Makefile

wfleming:
This could be used almost without changes in most of our engines: just change the value of IMAGE_NAME.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,9 @@
+.PHONY: image test
+
+IMAGE_NAME ?= codeclimate/codeclimate-duplication
+
+image:
+	docker build --rm -t $(IMAGE_NAME) .
+
+test: image
+	docker run --rm $(IMAGE_NAME) bundle exec rake
--- a/bin/test
+++ b/bin/test
@@ -1,5 +1,0 @@
-#!/bin/sh
-set -ex
-
-docker build -t codeclimate/codeclimate-duplication .
-docker run codeclimate/codeclimate-duplication bundle exec rake


Makefile
```diff
@@ -0,0 +1,11 @@
+.PHONY: all image test
+
+IMAGE_NAME=codeclimate/codeclimate-duplication
+
+all: image test
```
pbrisbin: Why would I `make all` when `make test` is more intention-revealing and has the same result?

Makefile
```diff
@@ -0,0 +1,11 @@
...
+
+image:
+	docker build -t $(IMAGE_NAME) .
```
pbrisbin: I always forget this too, but can you include `--rm` here. It prevents the intermediate `<none>` layers from building up (is my understanding).

Makefile
```diff
@@ -0,0 +1,11 @@
+.PHONY: all image test
+
+IMAGE_NAME=codeclimate/codeclimate-duplication
```
pbrisbin: I *think* the prevailing style is spaces around `=`. Also, how about letting it be env-overridden?

```
IMAGE_NAME ?= codeclimate/codeclimate-duplication
```

pbrisbin:
FYI: I'm being nit-heavy here because I can see this pattern spreading to other engines so it's worth getting right and pinning down style IMO.

Makefile
wfleming: Does that affect caching, though?

Makefile
wfleming: This is a holdover from the one I wrote for builder, with the idea being that just typing `make` would build an image & also run tests to confirm goodness. You can, of course, run `make test` explicitly.

I'm fine with removing it & saying the default task should simply be building the image. I'll update builder as well if that's preferred, since whichever way we do it should be consistent.

Actually, I'll go ahead and remove it in both places: earlier Makefiles (app, toolbox) don't do this.

Makefile
pbrisbin: Not to my knowledge. I'm not 100% clear on the details, but I know if I `docker build --rm ... && docker build --rm` the second is still all Cached steps.

wfleming: Ok, thanks.

wfleming:
Nits are good! I do expect whatever we do here to become the template for other engines, so it's worth swearing the details now so we're happy with it.

Makefile
```diff
@@ -0,0 +1,11 @@
...
+
+all: image test
+
+image:
```
wfleming: I'm not sure I want to keep this task name: as we work with more open source tools as "first party" engines, it'll become more likely those other projects will already have makefiles, so I think "image" might be too generic. I'm wondering if we should standardize on something more explicit that we can expect to use without issue everywhere. Maybe "codeclimate_engine"? Thoughts on this @pbrisbin?

pbrisbin: Hmm. It's worth questioning, but I'm not sure I share your concern:

- An existing project with an existing Makefile is unlikely to be using `image`, IMHO. While it is generic, I've never seen `make image` in the wild
- The intent of this target is (to me) not specifically to "make a codeclimate engine", but to build a docker image named `IMAGE_NAME` -- `make image` seems very apt

For those reasons, I'd be OK going with this for now. Could be worth sleeping on it though.

wfleming:
:eyes: @codeclimate/review

pbrisbin:
LGTM -- so sleek!

PR #55, opened 2015-12-17
rescue less

wfleming:
The Analyzer `#run` loop was rescuing *all* exceptions, which is clearly
excessive. But I don't think we should hard-fail on all exceptions, either:
the Ruby parser gem is definitely known to barf on some valid but
esoteric Ruby code, and I wouldn't be surprised if the other parsers had
similar edge cases.

So this changes behavior to only skip files for a known set of catchable
errors. The out-of-process parsers are a little tricky since all you can
get from them is an exit code and output streams: for now wrapping any non-zero exit from them
in our own exception class seems reasonable. They're not effected by Java heap problems, and since we don't control the heap anymore, any potential memory problems should kill the whole container with OOM.

I'm still catching all exceptions, but only so that we can log a message
about which file is impacted before we re-raise. Since a raw exception
that will cause an abort may not contain helpful information about the file
that triggered it, this would be helpful for debugging cases.

Note that there is a similar case of an excessive `rescue` in the `FileThreadPool`. I'm investigating & fixing that separately.

Thoughts, @codeclimate/review ?

### DIFF
--- a/lib/cc/engine/analyzers/analyzer_base.rb
+++ b/lib/cc/engine/analyzers/analyzer_base.rb
@@ -1,16 +1,27 @@
+require "cc/engine/analyzers/parser_error"
+
 module CC
   module Engine
     module Analyzers
       class Base
+        RESCUABLE_ERRORS = [
+          ::CC::Engine::Analyzers::ParserError,
+          ::Racc::ParseError,
+          ::Timeout::Error,
+        ]
+
         def initialize(engine_config:)
           @engine_config = engine_config
         end
 
         def run(file)
           process_file(file)
+        rescue *RESCUABLE_ERRORS => ex
+          $stderr.puts("Skipping file #{file} due to exception:")
+          $stderr.puts("(#{ex.class}) #{ex.message} #{ex.backtrace.join("\n")}")
         rescue => ex
-          $stderr.puts "Skipping file #{file} due to exception"
-          $stderr.puts "(#{ex.class}) #{ex.message} #{ex.backtrace.join("\n")}"
+          $stderr.puts("#{ex.class} error occurred processing file #{file}: aborting.")
+          raise ex
         end
 
         def files
--- a/lib/cc/engine/analyzers/analyzer_base.rb
+++ b/lib/cc/engine/analyzers/analyzer_base.rb
@@ -7,7 +7,6 @@ module CC
         RESCUABLE_ERRORS = [
           ::CC::Engine::Analyzers::ParserError,
           ::Racc::ParseError,
-          ::Timeout::Error,
         ]
 
         def initialize(engine_config:)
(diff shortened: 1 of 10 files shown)


lib/cc/engine/analyzers/command_line_runner.rb
```diff
@@ -0,0 +1,45 @@
...
+
+              exit_code = wait_thr.value
+
+              output = stdout.gets
+              stdout.close
+
+              err_output = stderr.gets
+              stderr.close
+
+              if 0 == exit_code
```
dblandin: I noticed that there was a `exit_code.to_i` before the move. Is that still needed here?

dblandin:
This looks like a great approach :+1:

lib/cc/engine/analyzers/analyzer_base.rb
```diff
@@ -1,16 +1,26 @@
+require "cc/engine/analyzers/parser_error"
+
 module CC
   module Engine
     module Analyzers
       class Base
+        RESCUABLE_ERRORS = [
```
wfleming: Opinion question for the team: should we also catch `Timeout::Error`? Flay does internally (not in a code path we're using, FWIW). I think the thinking behind it was that if the parser took too long on a file, it's probably a parser bug (or maybe an absurdly large file). We were implicitly catching it before.

lib/cc/engine/analyzers/command_line_runner.rb
wfleming: Nope. Different invocation. This is already an int.

dblandin: :+1:

lib/cc/engine/analyzers/analyzer_base.rb
dblandin: I'd try to match the behavior of Flay run on the command line then. If the Flay catches the timeout error, logs a skip, and continues on, I think we should do the same.

pbrisbin: @dblandin 's reasoning sounds good to me.

wfleming: I agree as well. It has been added now.

wfleming: Actually, I'm going to back this out: I remembered why I was hesitant about this in the first place. This series of changes was prompted by concerns of non-deterministic results from this engine (due to memory settings, but really a variety of system issues could come into play).

These timeouts are, by their nature, slightly non-deterministic. a file complex & big enough to be borderline might timeout some times & not others. Memory pressure on the system could cause execution time to vary between one run & another. Etc.

So instead I think I'm going to consider the timeout errors fatal & up the timeout limit to something absurd like 5 minutes. That should ensure it only gets triggered in pathological cases (like a parser bug leading to an infinite loop or something).

@jpignata this is relevant to your interests, so your thoughts are welcome.

PR #56, opened 2015-12-17
Don't rescue ThreadError in FileThreadPool

wfleming:
The previous implementation was blindly catching all ThreadError
exceptions, because that's what `#pop(true)` raises when the queue is
empty. However, there are other situations that would raise the same
class that we definitely shouldn't be silently ignoring.

This changes the code to actually check if the queue is empty before
popping, and removes the rescue entirely: if exceptions happen, we want
to know.

:eyes: @codeclimate/review

### DIFF
--- a/lib/cc/engine/analyzers/file_thread_pool.rb
+++ b/lib/cc/engine/analyzers/file_thread_pool.rb
@@ -17,11 +17,8 @@ module CC
 
           @workers = thread_count.times.map do
             Thread.new do
-              begin 
-                while item = queue.pop(true)
-                  yield item
-                end
-              rescue ThreadError
+              while !queue.empty? && (item = queue.pop(true))
+                yield item
               end
             end
           end


lib/cc/engine/analyzers/file_thread_pool.rb
```diff
@@ -17,11 +17,8 @@ def run(&block)
 
           @workers = thread_count.times.map do
             Thread.new do
-              begin 
-                while item = queue.pop(true)
-                  yield item
-                end
-              rescue ThreadError
+              while !queue.empty? && item = queue.pop(true)
```
dblandin: What do you think of using `#any?` here? If CC were running on this repo, I think it would recommend wrapping the assignment.

```ruby
while queue.any? && (item = queue.pop(true))
```

dblandin:
Just a couple nits. Looks good :+1:

lib/cc/engine/analyzers/file_thread_pool.rb
wfleming: The parens are a very good point. But there is no `#any?` defined on `Queue`[http://ruby-doc.org/core-2.2.3/Queue.html].

dblandin: Ah, so it doesn't. Thought it might mixin Enumerable, but it doesn't.

PR #57, opened 2015-12-18
Set Java Heap settings

wfleming:
Starting heap 512mb, max heap 1.25GB.

:eyes: @codeclimate/review

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,4 +25,7 @@ RUN npm install
 
 USER app
 
+# starting heap, max heap
+ENV JAVA_OPTS="-Xms512m -Xmx1280m"
+
 CMD ["/usr/src/app/bin/duplication"]


wfleming:
Sigh. JVM behavior can be strange & confusing.

Setting only the initial heap failed tests because it was larger than the default max heap. Which is strange because docs say it's either 64MB (really?) or 1GB, but you can also query for it at runtime, which I did in container and the value was, I think, Java's max int value. So I think we were really getting getting the 64mb max heap, which is too small.

Dockerfile
```diff
@@ -25,4 +25,6 @@ RUN npm install
 
 USER app
 
+ENV JAVA_OPTS="-Xms512m -Xmx1536m"
```
GordonDiggs: Maybe add a comment above this of what each of these do?

GordonDiggs:
One comment but otherwise LGTM

PR #58, opened 2015-12-18
set rubocop settings to our standard style

wfleming:
Updating these while convenient, and also since we haven't had PR integration turned for a while to check that's working.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -51,6 +51,11 @@ Style/GlobalVars:
   - "$mongo"
   - "$rollout"
 
+# Allow $! in config/initializers
+Style/SpecialGlobalVars:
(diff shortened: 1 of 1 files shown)


pbrisbin:
Should you just copy in the one from the style guide?

wfleming:
> Should you just copy in the one from the style guide?

I did?

pbrisbin:
Oh, the diff was so small, I mistook it for a modification of a `codeclimate --init`ed one. Not just syncing up with the style guide.

LGTM.

wfleming:
@pbrisbin I think I had copied in the styleguide one a while ago, so this was just a sync of changes since then. Hence the small diff.

PR #59, opened 2015-12-18
catch ENOENT

wfleming:
These can occur because of, for example, dangling symlinks, which are
the engine's responsibility.

:eyes: @codeclimate/review


wfleming:
Going to merge on green unless there are objections.

dblandin:
LGTM

PR #61, opened 2015-12-29
Tune remediation points to match Classic

ABaldwinHunter:
WIP - opening early for feedback

Changes include:
- Increase `base_points` for all four languages
- Add `points_per` value to each language
- change calculation formula from
    `base_points * issue.mass_threshold`
   to
   `base_points + overage * points_per`

Reference to rubric in classic:
https://github.com/codeclimate/analyzer/blob/master/lib/quality/rubric.rb#L12c

Note: python is handled separately in classic. It's not included in
rubric file above ^.

### DIFF
--- a/lib/cc/engine/analyzers/analyzer_base.rb
+++ b/lib/cc/engine/analyzers/analyzer_base.rb
@@ -13,7 +13,7 @@ module CC
 
         DEFAULT_MASS_THRESHOLD = 28
         BASE_POINTS = 1_500_000
-        POINTS_PER = 50_000
+        POINTS_PER_OVERAGE = 50_000
 
         def initialize(engine_config:)
           @engine_config = engine_config
@@ -41,8 +41,8 @@ module CC
           self.class::BASE_POINTS
         end
 
-        def points_per
-          self.class::POINTS_PER
+        def points_per_overage
+          self.class::POINTS_PER_OVERAGE
         end
 
         private
--- a/lib/cc/engine/analyzers/analyzer_base.rb
+++ b/lib/cc/engine/analyzers/analyzer_base.rb
@@ -11,6 +11,10 @@ module CC
           ::RubyParser::SyntaxError,
         ]
 
+        DEFAULT_MASS_THRESHOLD = 28
+        BASE_POINTS = 1_500_000
+        POINTS_PER = 50_000
+
         def initialize(engine_config:)
           @engine_config = engine_config
         end
@@ -35,6 +39,10 @@ module CC
 
         def base_points
           self.class::BASE_POINTS
+        end
+
+        def points_per
+          self.class::POINTS_PER
         end
 
         private
--- a/lib/cc/engine/analyzers/javascript/main.rb
+++ b/lib/cc/engine/analyzers/javascript/main.rb
@@ -15,8 +15,6 @@ module CC
             "**/*.jsx"
           ]
           LANGUAGE = "javascript"
-          DEFAULT_MASS_THRESHOLD = 40
-          BASE_POINTS = 3000
 
           private
 
--- a/lib/cc/engine/analyzers/reporter.rb
+++ b/lib/cc/engine/analyzers/reporter.rb
@@ -60,19 +60,15 @@ module CC
 
         attr_reader :engine_config, :language_strategy, :io
 
-        def mass_threshold
-          @mass_threshold ||= language_strategy.mass_threshold
-        end
-
         def new_violation(issue)
           hashes = flay.hashes[issue.structural_hash]
-          Violation.new(language_strategy.base_points, issue, hashes)
+          Violation.new(language_strategy, issue, hashes)
         end
 
         def flay_options
           {
             diff: false,
-            mass: mass_threshold,
+            mass: language_strategy.mass_threshold,
             summary: false,
             verbose: false,
             number: true,
--- a/lib/cc/engine/analyzers/violation.rb
+++ b/lib/cc/engine/analyzers/violation.rb
@@ -11,9 +11,7 @@ module CC
         DEFAULT_POINTS = 1_500_000
 
         def initialize(language, issue, hashes)
-          @base_points = language.base_points
-          @points_per = language.points_per
-          @threshold = language.mass_threshold
+          @language = language
           @issue = issue
           @hashes = hashes
         end
@@ -38,7 +36,7 @@ module CC
 
         def calculate_points
           if issue.mass >= threshold
-            base_points + (overage * points_per)
+            base_points + (overage * points_per_overage)
           else
             DEFAULT_POINTS
           end
@@ -46,7 +44,19 @@ module CC
 
         private
 
-        attr_reader :base_points, :points_per, :threshold, :hashes
+        attr_reader :language, :hashes
+
+        def base_points
+          language.base_points
+        end
+
+        def points_per_overage
+          language.points_per_overage
+        end
+
+        def threshold
+          language.mass_threshold
+        end
 
         def current_sexp
           @location ||= sorted_hashes.first
--- a/lib/cc/engine/analyzers/violation.rb
+++ b/lib/cc/engine/analyzers/violation.rb
@@ -8,8 +8,12 @@ module CC
       class Violation
         attr_reader :issue
 
-        def initialize(base_points, issue, hashes)
-          @base_points = base_points
+        DEFAULT_POINTS = 1_500_000
+
+        def initialize(language, issue, hashes)
+          @base_points = language.base_points
+          @points_per = language.points_per
+          @threshold = language.mass_threshold
           @issue = issue
           @hashes = hashes
         end
@@ -32,9 +36,17 @@ module CC
           "#{current_sexp.file}-#{current_sexp.line}"
         end
 
+        def calculate_points
+          if issue.mass >= threshold
+            base_points + (overage * points_per)
+          else
+            DEFAULT_POINTS
+          end
+        end
+
         private
 
-        attr_reader :base_points, :hashes
+        attr_reader :base_points, :points_per, :threshold, :hashes
 
         def current_sexp
           @location ||= sorted_hashes.first
@@ -56,8 +68,8 @@ module CC
           end
         end
 
-        def calculate_points
-          base_points * issue.mass
+        def overage
+          issue.mass - threshold
         end
 
         def format_location
--- a/spec/cc/engine/analyzers/violation_spec.rb
+++ b/spec/cc/engine/analyzers/violation_spec.rb
@@ -4,7 +4,7 @@ RSpec.describe CC::Engine::Analyzers::Violation, in_tmpdir: true do
   describe "#calculate_points" do
     context "when issue mass exceeds threshold" do
       it "calculates mass overage points" do
-        language = stub_language(base_points: 100, points_per: 5, mass_threshold: 10)
+        language = double(:language, base_points: 100, points_per_overage: 5, mass_threshold: 10)
         issue = double(:issue, mass: 15)
         hashes = []
 
@@ -19,7 +19,7 @@ RSpec.describe CC::Engine::Analyzers::Violation, in_tmpdir: true do
 
     context "when issue mass is less than threshold" do
       it "uses default" do
-        language = stub_language(base_points: 100, points_per: 5, mass_threshold: 18)
+        language = double(:language, base_points: 100, points_per_overage: 5, mass_threshold: 18)
         issue = double(:issue, mass: 15)
         hashes = []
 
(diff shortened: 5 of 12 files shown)


ABaldwinHunter:
@codeclimate/review Ready for review!

Fixed specs and lowered python threshold to 28 to match classic. 

With these tuning changes and those to `codeclimate-rubocop`, `app` is a 3.31.

See updated commit message for full summary of changes. 

Reference to rubric in classic for PHP, Ruby and JavaScript:
https://github.com/codeclimate/analyzer/blob/master/lib/quality/rubric.rb#L12c

And python:
https://github.com/codeclimate/analyzer/blob/master/lib/quality/profile.rb#L41

lib/cc/engine/analyzers/javascript/main.rb
```diff
@@ -15,8 +15,9 @@ class Main < CC::Engine::Analyzers::Base
             "**/*.jsx"
           ]
           LANGUAGE = "javascript"
-          DEFAULT_MASS_THRESHOLD = 40
-          BASE_POINTS = 3000
+          DEFAULT_MASS_THRESHOLD = 28
```
GordonDiggs: Would it work to define the ones that are the same across the board in `Base`? 

```ruby
[1] pry(main)> class Base
[1] pry(main)*   FOO = 5
[1] pry(main)*   
[1] pry(main)*   def foo
[1] pry(main)*     self.class::FOO
[1] pry(main)*   end  
[1] pry(main)* end  
=> :foo
[2] pry(main)> class A < Base
[2] pry(main)* end  
=> nil
[3] pry(main)> A.new.foo
=> 5
```

lib/cc/engine/analyzers/violation.rb
```diff
@@ -57,7 +61,15 @@ def name
         end
 
         def calculate_points
-          base_points * issue.mass
+          if issue.mass > threshold
+            base_points + overage * points_per
```
GordonDiggs: Might be good to wrap `overage * points_per` in parentheses to make this visually clearer

lib/cc/engine/analyzers/violation.rb
```diff
@@ -57,7 +61,15 @@ def name
         end
 
         def calculate_points
-          base_points * issue.mass
+          if issue.mass > threshold
```
wfleming: I think this should be `>=`? I think something that exactly hit the mass threshold is still an issue.

lib/cc/engine/analyzers/violation.rb
```diff
@@ -57,7 +61,15 @@ def name
...
+            base_points + overage * points_per
+          else
```
wfleming: This seems unnecessary (`DEFAULT_POINTS` seems unnecessary in general). If the issue mass is not greater or equal than the threshold, it's not an issue at all, and doesn't need points. Is this maybe vestigial from how classic decided if something was worth reporting?

lib/cc/engine/analyzers/javascript/main.rb
ABaldwinHunter: sure that sounds good. 

In general, it felt to me like there were opportunities to refactor in this code, but I wanted to avoid additional changes in this PR and refactor separately.

I'm happy with the threshold one though, adding that now!

lib/cc/engine/analyzers/violation.rb
```diff
@@ -57,7 +61,15 @@ def name
         end
 
         def calculate_points
```
wfleming: This new implementation seems involved enough to be worth speccing. Seems like it might be time to create a spec for this class.

lib/cc/engine/analyzers/violation.rb
ABaldwinHunter: Thought about that. Was on fence but now settled and agree.

Making change - thanks!

lib/cc/engine/analyzers/violation.rb
ABaldwinHunter: Thanks, I didn't realize that. Making change.

Anyway, interestingly, since the `DEFAULT_POINTS` = `1_500_000`, it's the same result on either leg of conditional. 

But your suggestion is much clearer :) thanks

lib/cc/engine/analyzers/reporter.rb
```diff
@@ -64,9 +64,17 @@ def mass_threshold
...
+        end
+
+        def points_per
+          @points_per ||= language_strategy.points_per
+        end
+
         def new_violation(issue)
           hashes = flay.hashes[issue.structural_hash]
-          Violation.new(language_strategy.base_points, issue, hashes)
+          Violation.new(base_points, points_per, mass_threshold, issue, hashes)
```
wfleming: Since all of these come directly from `language_strategy`, would it be better to just pass the language strategy to `Violation`? The code seems to be saying is that a violation's properties are fairly dependent on the specific language it's from, so that doesn't unreasonable to me.

lib/cc/engine/analyzers/violation.rb
wfleming: That's partly why I pointed it out: `DEFAULT_POINTS` *happens* to be correct right now, but only because `BASE_POINTS` is the same for all the languages. Since that could change, depending upon it would basically be a bug-in-waiting.

As @GordonDiggs suggested above, if we could define those values in the `Base` impl, that would certainly help clarify things.

lib/cc/engine/analyzers/reporter.rb
ABaldwinHunter: @wfleming Yeah, I think I agree with you.

However, I think that's a separate refactor PR. I think this PR should stay focused on the points changes to minimize impact, but open to other thoughts.

lib/cc/engine/analyzers/analyzer_base.rb
```diff
@@ -11,6 +11,10 @@ class Base
           ::RubyParser::SyntaxError,
         ]
 
+        DEFAULT_MASS_THRESHOLD = 28
+        BASE_POINTS = 1_500_000
+        POINTS_PER = 50_000
```
ABaldwinHunter: @GordonDiggs I'm slightly worried that using inferred constants makes the code harder to follow for future devs. Now that you see it, any doubts?

If you're not worried, I'm not.

GordonDiggs: I don't think it's any more confusing than using the inherited methods, which I don't find confusing at all.

lib/cc/engine/analyzers/violation.rb
GordonDiggs: Agreed

http://thumbstir.com/jenga-cat-is-back-and-better-than-ever/2901

lib/cc/engine/analyzers/analyzer_base.rb
ABaldwinHunter: Okay cool. Thanks!

ABaldwinHunter:
@codeclimate/review Ready for re-review!

Updates: 
- Refactor `Violation` class to initialize with `language_strategy` obj (fewer args)*
- Add calculation specs

* thanks @dblandin / @wfleming for suggestion

ABaldwinHunter:
Punt on `#format` spec because already tested by `main_spec`s and a bit hairy for integration.

I'd be interested in doing some additional refactoring of existing duplication code in a separate PR to follow.

lib/cc/engine/analyzers/analyzer_base.rb
dblandin: What do you think of `BASE_POINTS` and `MASS_OVERAGE_POINTS` instead of `POINTS_PER`?

I didn't find the current name intuitive.

spec/cc/engine/analyzers/violation_spec.rb
```diff
@@ -0,0 +1,59 @@
+RSpec.describe CC::Engine::Analyzers::Violation, in_tmpdir: true do
+  include AnalyzerSpecHelpers
+
+  describe "#calculate_points" do
+    context "when issue mass exceeds threshold" do
+      it "calculates mass overage points" do
+        language = stub_language(base_points: 100, points_per: 5, mass_threshold: 10)
```
dblandin: What do you think of changing the name to `strategy` and inlining instead of extracting a method?

```ruby
strategy = double(base_points: 100, points_per: 5, mass_threshold: 10)

```

lib/cc/engine/analyzers/violation.rb
```diff
@@ -8,8 +8,12 @@ module Analyzers
...
         attr_reader :issue
 
-        def initialize(base_points, issue, hashes)
-          @base_points = base_points
+        DEFAULT_POINTS = 1_500_000
+
+        def initialize(language, issue, hashes)
+          @base_points = language.base_points
+          @points_per = language.points_per
+          @threshold = language.mass_threshold
```
dblandin: I think it's fine to assign the whole strategy here and call methods on the strategy elsewhere in the class.

dblandin:
A few nits. Looking good :+1:

spec/cc/engine/analyzers/violation_spec.rb
ABaldwinHunter: I like that much better. Thanks!

lib/cc/engine/analyzers/violation.rb
ABaldwinHunter: Good call

lib/cc/engine/analyzers/analyzer_base.rb
```diff
@@ -11,6 +11,10 @@ class Base
           ::RubyParser::SyntaxError,
         ]
 
+        DEFAULT_MASS_THRESHOLD = 28
```
wfleming: I want to clarify that this is intentional and vetted: previously these were different between PHP & Python/Javascript, and this change makes them the same, but keeps Ruby different.

The previous values were, I believe, picked after a lot of manual testing to get something that felt "right". What the "right" default threshold choice is for a language is dependent on how the parser the engines uses for that language ends up expressing ASTs. I.e. small parser differences can change how verbose a mass threshold ends up seeming.

Based on [the classic code](https://github.com/codeclimate/analyzer/blob/master/lib/quality/rubric.rb#L12c), it looks like this value was taken directly from there (and it didn't apply to Python in classic). I'd personally be pretty surprised if the AST results of all the languages was the same between classic & this engine: at the very least I'm pretty sure the JS parser behavior is likely to be different.

For that reason, I'm not sure the classic thresholds will make sense here, because they are likely to produce different results here than they did in classic. Do you have a sample corpus you've been using to verify same results for each of these languages between classic/these changes?

lib/cc/engine/analyzers/analyzer_base.rb
ABaldwinHunter: That sounds okay to me.  Only arguments to consider against:

1) The `per` naming is consistent with some of our usages in engines and classic:
classic: [cost_per](https://github.com/codeclimate/analyzer/blob/master/lib/quality/rubric.rb#L5)
eslint: [cost_per](https://github.com/codeclimate/codeclimate-eslint/blob/master/lib/checks.js#L120)
rubocop: [per_violation_points](https://github.com/codeclimate/codeclimate-rubocop/blob/master/lib/cc/engine/violation_decorator.rb#L33)

2) the total amount of `overage points` is really this number times the overage (mass - threshold).

lib/cc/engine/analyzers/analyzer_base.rb
ABaldwinHunter: @wfleming Thanks for the tip about the thought behind the thresholds and parsing. I can investigate that. 

I reference the classic rubric you mentioned in [my commit](https://github.com/codeclimate/codeclimate-duplication/pull/61/commits), along with the Python source.

I've been keeping track of test repos [here](https://gist.github.com/ABaldwinHunter/e280310b4987766efd8c).

In general, it seemed that on classic, duplication and complexity issues had much greater grade impact than they did on Platform.

I tested the current duplication setup on `app` - but could do some further testing with this exact setup on additional repos.

lib/cc/engine/analyzers/analyzer_base.rb
ABaldwinHunter: Perhaps `points_per_overage`?

dblandin: That works for me :+1:

GordonDiggs:
This looks pretty great to me pending green builds

lib/cc/engine/analyzers/analyzer_base.rb
wfleming: I guess it's a question of effort vs accuracy, but if the intent is to get results similar to classic, we should probably have a small corpus of files for each language, get the overall mass of each file as calculated both by classic & by this engine, and compare those to see if the thresholds (and to a lesser extent `per_point` values) need to be scaled to get matching results.

ABaldwinHunter:
@wfleming Thanks for feedback. After posting a question in dev, I might break this change up, and just ship the ruby update. 

I did some additional screening on our own (ruby) repos and those grades look consistent with Classic.
 
I hadn't realized that the settings in Classic wouldn't translate directly for the other languages. 
In preliminary testing, flask gets: a `2.7` on engines now, `2.38` on classic, `1.93` with point updates here only, `1.7` with point updates and threshold change.
 
Sounds like it requires more data gathering to assess, and determine desired grades.

On the positive side: users have some flexibility to configure their own thresholds in `.codeclimate.yml`.

ABaldwinHunter:
Closing in favor of piecemeal approach, starting with Ruby: https://github.com/codeclimate/codeclimate-duplication/pull/67/files

PR #62, opened 2015-12-29
Catch RubyParser exceptions

wfleming:
Depending on what causes RubyParser to fail, the exception could be from
Racc or from RubyParser, so we should catch these as well.

@codeclimate/review @abaldwinhunter

### DIFF
--- a/lib/cc/engine/analyzers/analyzer_base.rb
+++ b/lib/cc/engine/analyzers/analyzer_base.rb
@@ -8,6 +8,7 @@ module CC
           ::CC::Engine::Analyzers::ParserError,
           ::Errno::ENOENT,
           ::Racc::ParseError,
+          ::RubyParser::SyntaxError,
         ]
 
         def initialize(engine_config:)


GordonDiggs:
LGTM. Funny that CC failed with this

ABaldwinHunter:
Nice

wfleming:
Um...yeah, I'm trying to think of how to explain the CC failure before I merge because it concerns me.

CC was green on #60, which was the previous version of the duplication engine (b226). The later builds on #61 and #62 have been the newest duplication engine release (b258), and they've failed on parser errors. (On the same commit, even: compare [this build](https://codeclimate.com/github/codeclimate/codeclimate-duplication/builds/18) and [this build](https://codeclimate.com/github/codeclimate/codeclimate-duplication/builds/19)).

Gem versions have not been changed in months, so they *should* be identical, and parser results shouldn't vary. And the changes in #60 aren't even involved with Ruby parsing at all (that happens in-line, unlike other languages that shell out to their respective parsers).

So that's weird & I don't know how to explain it yet.

wfleming:
Ugh. I've explained it: the prior duplication engine (b226) was still the very old one that just swallowed every exception ever. So of course it didn't fail.

PR #63, opened 2015-12-29
skip file message: format everything in one call

wfleming:
I noticed that some logs reporting skip file messages seemed to
interleave multiple messages: That's not terribly surprising due to
concurrency.

This changes the message formatting a bit & does it all in one method
call to avoid the threading issues.

@codeclimate/review


GordonDiggs:
:+1:

PR #65, opened 2015-12-30
Catch RuntimeError when parsing

wfleming:
RubyParser raises `RuntimeError` instead of anything more specific in
certain parse error contexts.

This presents a challenge for catching these, since many error types
(including `Timeout::Error`, which we don't want to catch) inherit from
`RuntimeError`. So I've had to shift from two `rescue` clauses to one
`rescue` clause that uses internal logic to decide if the error should
be re-raised.

Updated follow-up to #64, @codeclimate/review.

### DIFF
--- a/lib/cc/engine/analyzers/analyzer_base.rb
+++ b/lib/cc/engine/analyzers/analyzer_base.rb
@@ -9,7 +9,8 @@ module CC
           ::Errno::ENOENT,
           ::Racc::ParseError,
           ::RubyParser::SyntaxError,
-        ]
+          ::RuntimeError,
+        ].freeze
 
         def initialize(engine_config:)
           @engine_config = engine_config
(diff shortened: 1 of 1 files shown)


lib/cc/engine/analyzers/analyzer_base.rb
```diff
@@ -17,11 +18,13 @@ def initialize(engine_config:)
 
         def run(file)
           process_file(file)
-        rescue *RESCUABLE_ERRORS => ex
-          $stderr.puts("Skipping file #{file} due to exception (#{ex.class}): #{ex.message}\n#{ex.backtrace.join("\n")}")
         rescue => ex
-          $stderr.puts("#{ex.class} error occurred processing file #{file}: aborting.")
-          raise ex
+          if RESCUABLE_ERRORS.map { |klass| ex.instance_of?(klass) }.include?(true)
```
wfleming: Before anybody asks: `ActiveSupport` is not included in this gem, and it doesn't seem worth adding just for this.

ABaldwinHunter:
LGTM

PR #66, opened 2015-12-30
bump concurrent-ruby to release version

wfleming:
We were on a pre-release version of this gem, and there's a stable
release now.

cc @codeclimate/review


ABaldwinHunter:
LGTM

PR #67, opened 2015-12-30
Tune RUBY remediation points to match Classic

ABaldwinHunter:
Changes include:
- Refactor `calculate_points` method to be purview of language,
not `Violation` - to permit independent calculation formulas
- Increase ruby `BASE_POINTS`
- Change calculation formula from
    - `base_points * issue.mass_threshold`
  to
   - `base_points + overage * points_per`

Rubric in classic for Ruby:
https://github.com/codeclimate/analyzer/blob/master/lib/quality/rubric.rb#L12c

Updates for other languages to come post additional language-specific tuning

replaces https://github.com/codeclimate/codeclimate-duplication/pull/61/files
@codeclimate/review 
@brynary

### DIFF
--- a/lib/cc/engine/analyzers/violation.rb
+++ b/lib/cc/engine/analyzers/violation.rb
@@ -8,8 +8,8 @@ module CC
       class Violation
         attr_reader :issue
 
-        def initialize(base_points, issue, hashes)
-          @base_points = base_points
+        def initialize(language_strategy, issue, hashes)
+          @language_strategy = language_strategy
           @issue = issue
           @hashes = hashes
         end
@@ -34,7 +34,7 @@ module CC
 
         private
 
-        attr_reader :base_points, :hashes
+        attr_reader :language_strategy, :hashes
 
         def current_sexp
           @location ||= sorted_hashes.first
@@ -57,7 +57,7 @@ module CC
         end
 
         def calculate_points
-          base_points * issue.mass
+          language_strategy.calculate_points(issue)
         end
 
         def format_location
--- a/spec/cc/engine/analyzers/ruby/main_spec.rb
+++ b/spec/cc/engine/analyzers/ruby/main_spec.rb
@@ -4,60 +4,105 @@ require 'cc/engine/analyzers/file_list'
 require 'flay'
 require 'tmpdir'
 
-RSpec.describe CC::Engine::Analyzers::Ruby::Main, in_tmpdir: true do
-  include AnalyzerSpecHelpers
+module CC::Engine::Analyzers
+  RSpec.describe Ruby::Main, in_tmpdir: true do
+    include AnalyzerSpecHelpers
 
-  describe "#run" do
-    it "prints an issue" do
-      create_source_file("foo.rb", <<-EORUBY)
-          describe '#ruby?' do
-            before { subject.type = 'ruby' }
+    describe "#run" do
+      it "prints an issue" do
+        create_source_file("foo.rb", <<-EORUBY)
+            describe '#ruby?' do
+              before { subject.type = 'ruby' }
 
-            it 'returns true' do
-              expect(subject.ruby?).to be true
+              it 'returns true' do
+                expect(subject.ruby?).to be true
+              end
             end
-          end
 
-          describe '#js?' do
-            before { subject.type = 'js' }
+            describe '#js?' do
+              before { subject.type = 'js' }
 
-            it 'returns true' do
-              expect(subject.js?).to be true
+              it 'returns true' do
+                expect(subject.js?).to be true
+              end
             end
-          end
-      EORUBY
+        EORUBY
 
-      result = run_engine(engine_conf).strip
-      json = JSON.parse(result)
+        result = run_engine(engine_conf).strip
+        json = JSON.parse(result)
 
-      expect(json["type"]).to eq("issue")
-      expect(json["check_name"]).to eq("Similar code")
-      expect(json["description"]).to eq("Similar code found in 1 other location")
-      expect(json["categories"]).to eq(["Duplication"])
-      expect(json["location"]).to eq({
-        "path" => "foo.rb",
-        "lines" => { "begin" => 1, "end" => 5 },
-      })
-      expect(json["remediation_points"]).to eq(360000)
-      expect(json["other_locations"]).to eq([
-        {"path" => "foo.rb", "lines" => { "begin" => 9, "end" => 13} },
-      ])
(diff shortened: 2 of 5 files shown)


lib/cc/engine/analyzers/violation.rb
```diff
@@ -8,8 +8,8 @@ module Analyzers
       class Violation
         attr_reader :issue
 
-        def initialize(base_points, issue, hashes)
-          @base_points = base_points
+        def initialize(language, issue, hashes)
+          @language = language
```
dblandin: Would calling this either `strategy` or `language_strategy` better match the variable [passed in](https://github.com/codeclimate/codeclimate-duplication/pull/67/files#diff-d285873a9a1a70964311d140233c6dfaR65)?

spec/cc/engine/analyzers/ruby/main_spec.rb
```diff
@@ -57,6 +57,49 @@
     end
   end
 
+  describe "#calculate_points(issue)" do
+    let(:ruby) { CC::Engine::Analyzers::Ruby::Main.new(engine_config: engine_conf) }
+    let(:base_points) { CC::Engine::Analyzers::Ruby::Main::BASE_POINTS }
+    let(:points_per) { CC::Engine::Analyzers::Ruby::Main::POINTS_PER_OVERAGE }
+    let(:threshold) { CC::Engine::Analyzers::Ruby::Main::DEFAULT_MASS_THRESHOLD }
```
dblandin: You might want to wrap this whole spec with `module CC::Engine::Analyzers::Ruby`:

```ruby
module CC::Engine::Analyzers::Ruby
  describe Main, in_tmpdir: true do
    # ...

     describe "#calculate_points(issue)" do
        let(:ruby) { Main.new(engine_config: engine_conf) }
        let(:base_points) { Main::BASE_POINTS }
        let(:points_per) { Main::POINTS_PER_OVERAGE }
        let(:threshold) { Main::DEFAULT_MASS_THRESHOLD }
      end

      # ...
  end
end

File: spec/cc/engine/analyzers/ruby/main_spec.rb [hunk 3]
```diff
@@ -57,6 +57,49 @@
     end
   end
 
+  describe "#calculate_points(issue)" do
+    let(:ruby) { CC::Engine::Analyzers::Ruby::Main.new(engine_config: engine_conf) }
```
dblandin: What do you think of naming this `analyzer` instead of `ruby`?

lib/cc/engine/analyzers/violation.rb
ABaldwinHunter: I don't really understand the reasoning behind `language_strategy` instead of `language`?

dblandin:
@ABaldwinHunter This change looks good to me. Easy to follow :+1:

Just a few minor nits.

Are we intending to support independent point calculations in the long term or will we eventually standardize cross-language with `base_points + (overage * points_per_overage)`?

dblandin:
One other question about your PR description. Is the current calculation `base_points * issue.mass_threshold` or `base_points * issue.mass`? Seems to be the ladder from [this deletion](https://github.com/codeclimate/codeclimate-duplication/pull/67/files#diff-cb10d1962a1e9ec298481b1f291a09d2L60).

spec/cc/engine/analyzers/ruby/main_spec.rb
ABaldwinHunter: :+1:  thanks

lib/cc/engine/analyzers/violation.rb
dblandin: The main issue for me is the inconsistency. We're passing in an object named `language_strategy` in `lib/cc/engine/analyzers/reporter.rb` but assigning it internally within Violation as `language`. We should probably stick with one name.

It seems like what we're passing in, based on the methods called on the object, is more like a strategy (or configuration) than a "language".

spec/cc/engine/analyzers/ruby/main_spec.rb
ABaldwinHunter: :+1:

lib/cc/engine/analyzers/violation.rb
ABaldwinHunter: That is, I agree that `strategy` would be in keeping with the argument passed in, but I think the original name is a bit confusing.

ABaldwinHunter:
@dblandin Thanks for feedback!

I think we are going to move to `base_points + (overage * points_per)` for all languages v soon. 

For ruby, the formula is 
`base_points + overage * points_per`

`Ruby::Main` [overwrites](https://github.com/codeclimate/codeclimate-duplication/pull/67/files#diff-415d8ecf3775cdd67402dc77177f7b15R25) the inherited `calculate_points`.

lib/cc/engine/analyzers/violation.rb
dblandin: Hmmm, is this object always a subclass of module `CC::Engine::Analyzers::Base`?

ABaldwinHunter: Ah, missed your second comment before. 

I'm super loathe to use `strategy` but will change it to `language_strategy` if you feel strongly. :)

ABaldwinHunter:
Updates in place!

lib/cc/engine/analyzers/violation.rb
ABaldwinHunter: Now seeing that comment. Yes it is

dblandin: Then what do you think of using the name `analyzer` anywhere that `language` or `language_strategy` is currently used. That might clear up the confusion.

ABaldwinHunter: I think it's outside the scope of this PR and would advocate changing the naming in a separate refactor soon to follow. 

wdyt?

dblandin: Sounds good :+1:

spec/cc/engine/analyzers/ruby/main_spec.rb
```diff
@@ -4,60 +4,105 @@
...
+          expect(points).to eq(expected_points)
+        end
+      end
     end
-  end
 
-  def engine_conf
-    CC::Engine::Analyzers::EngineConfig.new({})
+    def engine_conf
+      EngineConfig.new({})
```
dblandin: This is probably small enough now to inline on L62.

dblandin:
Looking good :+1:

PR #68, opened 2016-01-03
Do not limit nesting depth of parsed JSON

wfleming:
I turned up some instances of files that killed the engine due to
deeply nested JSON results from the parser: the JSON gem by default puts
a limit on this.

:eyes: @codeclimate/review

### DIFF
--- a/lib/cc/engine/analyzers/parser_base.rb
+++ b/lib/cc/engine/analyzers/parser_base.rb
@@ -1,0 +1,13 @@
+module CC
+  module Engine
+    module Analyzers
+      class ParserBase
+        private
+
+        def parse_json(text)
+          JSON.parse(text, max_nesting: false)
+        end
+      end
+    end
+  end
+end
--- a/lib/cc/engine/analyzers/analyzer_base.rb
+++ b/lib/cc/engine/analyzers/analyzer_base.rb
@@ -1,4 +1,5 @@
 require "cc/engine/analyzers/parser_error"
+require "cc/engine/analyzers/parser_base"
 
 module CC
   module Engine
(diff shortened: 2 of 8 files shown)


brynary:
Historical context: In the past we've parsed with unlimited depth. Since this is in an engine, I think it might be worth doing that as the simplest possible thing (and not adding more exception handling, which can be troublesome) and seeing where that leaves us.

wfleming:
@brynary OK, I didn't realize we had an unlimited depth in classic analyzer. Will update.

GordonDiggs:
LGTM

lib/cc/engine/analyzers/parser_base.rb
```diff
@@ -0,0 +1,13 @@
+module CC
+  module Engine
+    module Analyzers
+      class ParserBase
+        protected
```
pbrisbin: I think this can be `private`. Ruby's method access is an odd-ball:

- private - cannot be called with a target (even `self.` will raise)
- protected - can only be called with a target if you're the same class or a subclass

  This is mostly (only?) useful for otherwise private attributes that need to invoked with a target as part of an equality check:

  ```rb
  class Foo
    def ==(other)
      private_thing == other.private_thing
    end
  end
  ```

  This won't work if `#private_thing` is actually `private`, so it would be `protected`.

pbrisbin:
One nit. LGTM

lib/cc/engine/analyzers/parser_base.rb
wfleming: I think I do this habitually more for how it reads than how it actually works. But good point, changing.

PR #71, opened 2016-01-04
PHP Parser: handle INF/NaN nodes

wfleming:
I tracked this back to this constant number in a source file.

It turns out the literal INF constant in a PHP source is fine:
the parser represents constants directly using the identifier, so
it's fine in JSON.

But the constant number is larger than PHP's max number, so it basically
overflows & get represented as INF after parsing.

On the belief that similar code might result in a NaN constant, I'm also
covering that one. This represents both values using constant strings
that are unlikely to actually appear in anyone's source code.

:eyes: @codeclimate/review


GordonDiggs:
LGTM

PR #76, opened 2016-01-15
Specify similar vs identical code in description

ABaldwinHunter:
This information is already present on issue check_name, but the description is more visible.
@codeclimate/review 

image

### DIFF
--- a/lib/cc/engine/analyzers/violation.rb
+++ b/lib/cc/engine/analyzers/violation.rb
@@ -17,7 +17,7 @@ module CC
         def format
           {
             "type": "issue",
-            "check_name": name,
+            "check_name": check_name,
             "description": description,
             "categories": ["Duplication"],
             "location": format_location,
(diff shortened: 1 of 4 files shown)


pbrisbin:
Nice. LGTM

ABaldwinHunter:
Fixing test failure

lib/cc/engine/analyzers/violation.rb
```diff
@@ -48,7 +48,7 @@ def other_sexps
           @other_locations ||= sorted_hashes.drop(1)
         end
 
-        def name
+        def check_name
```
GordonDiggs: Should identical vs. similar factor into the fingerprint? Maybe out of scope for this?

pbrisbin: Not sure, but I don't think so.

Issues that get better or worse don't change fingerprints. I think a duplication issue moving from Similar to Identical or back is related to, and part of, that issue getting better or worse.

PR #78, opened 2016-01-19
Tune python duplication remediation points

ABaldwinHunter:
- Reduce AST threshold from 40 to ~~31~~ 32 (classic is 28)
- update point formula to match classic computation

Change from
     `remediations_points = n * score`
to
    `remediation_points = x + (score-threshold) * y`

This change increases parity with classic and overall increases the number of duplication issues reported.

[link to grade comparisons](https://gist.github.com/ABaldwinHunter/e280310b4987766efd8c)

**Note**: @GordonDiggs and I are exploring exactly why there are parser differences between Classic and Platform - there are a number of relevant factors, including likely differences between versions of Python. 

Note on mass difference:

The mass of node corresponds to its size. Specifying a minimum threshold tells Code Climate to ignore duplication in nodes below a certain size (e.g. one liners).

The issue's Flay score is the result of its **mass** * **number of occurrences** (or number of occurrences ^ 2, if the code is identical).

Comparing issue **mass** between parser in Platform and Classic:

| Platform  | Classic |  Platform / Classic    |
------------ | --------------- | ------------
 42      |  39   |  1.07
 45      |  40   |  1.125
 66      |  57   |  1.15789
 123      |  109  | 1.1284
  126    |  93  | 1.3548
  246      |  218  | 1.1284

I've estimated the factor of mass difference to be ~ 1.15
Since the default Python duplication mass threshold on Classic was 28, and  `28 * 1.15 = 32.19999`, I've lowered our current default threshold for Python on Platform from 40 to 32.

On Classic, Python duplication issues were penalized in terms of remediation points as follows:

`1_500_000 + overage * 50_000`
where overage = **score** - **threshold**
( and score = f(mass) ) 

I've kept the base points but lowered the per_cost to 30_000 to account for the difference in mass parsing (which gets amplified in the points calculation. 

@codeclimate/review

### DIFF
--- a/config/contents/duplicated_code.md.erb
+++ b/config/contents/duplicated_code.md.erb
@@ -9,7 +9,7 @@ When you violate DRY, bugs and maintenance problems are sure to follow. Duplicat
 ## Issue Mass
 
 Duplicated code has a calculated mass, which can be thought of as a measure of how much logic has been duplicated.
-This issue has a mass of `<%= issue.mass %>`: if you would like to change the minimum mass that will be reported as an issue, please see the details in [`codeclimate-duplication`'s documentation](https://github.com/codeclimate/codeclimate-duplication).
+This issue has a mass of `<%= mass %>`: if you would like to change the minimum mass that will be reported as an issue, please see the details in [`codeclimate-duplication`'s documentation](https://github.com/codeclimate/codeclimate-duplication).
 
 ## Refactorings
 
--- a/lib/cc/engine/analyzers/issue.rb
+++ b/lib/cc/engine/analyzers/issue.rb
@@ -1,0 +1,89 @@
+require "cc/engine/analyzers/sexp_lines"
+require "cc/engine/analyzers/violation_read_up"
+require "digest"
+
+module CC
+  module Engine
+    module Analyzers
+      class Issue
+        def initialize(language_strategy:, check_name:, current_sexp:, other_sexps:)
+          @language_strategy = language_strategy
+          @check_name = check_name
+          @current_sexp = current_sexp
+          @other_sexps = other_sexps
+        end
+
+        def format # rubocop:disable Metrics/MethodLength
+          {
+            "type": "issue",
+            "check_name": check_name,
+            "description": description,
+            "categories": ["Duplication"],
+            "location": format_location,
+            "remediation_points": calculate_points,
+            "other_locations": format_other_locations,
+            "content": content_body,
+            "fingerprint": fingerprint,
+          }
+        end # rubocop:enable Metrics/MethodLength
+
+        def report_name
+          "#{current_sexp.file}-#{current_sexp.line}"
+        end
+
+        def mass
+          current_sexp.mass
+        end
+
+        private
+
+        attr_reader :language_strategy, :check_name, :other_sexps, :current_sexp
+
+        def calculate_points
+          language_strategy.calculate_points(mass)
+        end
+
+        def format_location
+          format_sexp(current_sexp)
+        end
+
+        def format_other_locations
+          other_sexps.map do |sexp|
+            format_sexp(sexp)
+          end
+        end
+
+        def format_sexp(sexp)
+          lines = SexpLines.new(sexp)
+          {
+            "path": sexp.file.gsub(%r{^./}, ""),
+            "lines": {
+              "begin": lines.begin_line,
+              "end": lines.end_line,
+            },
+          }
+        end
+
+        def content_body
+          { "body": ViolationReadUp.new(mass).contents }
+        end
+
+        def fingerprint
+          digest = Digest::MD5.new
+          digest << current_sexp.file
+          digest << "-"
+          digest << current_sexp.mass.to_s
+          digest << "-"
+          digest << check_name
+          digest.to_s
+        end
+
+        def description
+          description = "#{check_name} found in #{(other_sexps.length)} other location"
+          description += "s" if other_sexps.length > 1
+          description
+        end
+      end
+    end
+  end
+end
--- a/lib/cc/engine/analyzers/python/main.rb
+++ b/lib/cc/engine/analyzers/python/main.rb
@@ -12,10 +12,19 @@ module CC
         class Main < CC::Engine::Analyzers::Base
           LANGUAGE = "python"
           DEFAULT_PATHS = ["**/*.py"]
-          DEFAULT_MASS_THRESHOLD = 40
-          BASE_POINTS = 1000
+          DEFAULT_MASS_THRESHOLD = 32
+          BASE_POINTS = 1_500_000
+          POINTS_PER_OVERAGE = 50_000
+
+          def calculate_points(mass)
+            BASE_POINTS + (overage(mass) * POINTS_PER_OVERAGE)
+          end
 
           private
+
+          def overage(mass)
+            mass - mass_threshold
+          end
 
           def process_file(path)
             Node.new(::CC::Engine::Analyzers::Python::Parser.new(File.binread(path), path).parse.syntax_tree, path).format
--- a/lib/cc/engine/analyzers/reporter.rb
+++ b/lib/cc/engine/analyzers/reporter.rb
@@ -38,9 +38,11 @@ module CC
           flay.report(StringIO.new).each do |issue|
             violation = new_violation(issue)
 
-            unless reports.include?(violation.report_name)
-              reports.add(violation.report_name)
-              io.puts "#{violation.format.to_json}\0"
+            violation.occurrences.each do |occurrence|
+              unless reports.include?(occurrence.report_name)
+                reports.add(occurrence.report_name)
+                io.puts "#{occurrence.format.to_json}\0"
+              end
             end
           end
         end
--- a/lib/cc/engine/analyzers/violation.rb
+++ b/lib/cc/engine/analyzers/violation.rb
@@ -1,52 +1,24 @@
-require "cc/engine/analyzers/sexp_lines"
-require "cc/engine/analyzers/violation_read_up"
-require "digest"
+require "cc/engine/analyzers/issue"
 
 module CC
   module Engine
     module Analyzers
       class Violation
-        attr_reader :issue
-
         def initialize(language_strategy, issue, hashes)
           @language_strategy = language_strategy
-          @issue = issue
           @hashes = hashes
+          @issue = issue
         end
 
-        def format
-          {
-            "type": "issue",
-            "check_name": check_name,
-            "description": description,
-            "categories": ["Duplication"],
-            "location": format_location,
-            "remediation_points": calculate_points,
-            "other_locations": format_other_locations,
-            "content": content_body,
-            "fingerprint": fingerprint
-          }
-        end
-
-        def report_name
-          "#{current_sexp.file}-#{current_sexp.line}"
+        def occurrences
+          hashes.map.with_index do |sexp, i|
+            Issue.new(language_strategy: language_strategy, check_name: check_name, current_sexp: sexp, other_sexps: other_sexps(hashes.dup, i))
+          end
         end
 
         private
 
-        attr_reader :language_strategy, :hashes
-
-        def current_sexp
-          @location ||= sorted_hashes.first
-        end
-
-        def sorted_hashes
-          @_sorted_hashes ||= hashes.sort_by(&:file)
-        end
-
-        def other_sexps
-          @other_locations ||= sorted_hashes.drop(1)
-        end
+        attr_reader :language_strategy, :hashes, :issue
 
         def check_name
           if issue.identical?
@@ -56,53 +28,9 @@ module CC
           end
         end
 
-        def calculate_points
-          language_strategy.calculate_points(issue)
-        end
-
-        def format_location
-          format_sexp(current_sexp)
-        end
-
-        def format_other_locations
-          other_sexps.map do |sexp|
-            format_sexp(sexp)
-          end
-        end
-
-        def format_sexp(sexp)
-          lines = SexpLines.new(sexp)
-          {
-            "path": sexp.file.gsub(%r(^./), ""),
-            "lines": {
-              "begin": lines.begin_line,
-              "end": lines.end_line,
-            },
-          }
-        end
-
-        def content_body
-          @_content_body ||= { "body": ViolationReadUp.new(issue).contents }
-        end
-
-        def fingerprint
-          digest = Digest::MD5.new
-          digest << current_sexp.file
-          digest << "-"
-          digest << current_sexp.mass.to_s
-          digest << "-"
-          digest << occurrences.to_s
-          digest.to_s
-        end
-
-        def description
-          description = "#{check_name} found in #{occurrences} other location"
-          description += "s" if occurrences > 1
-          description
-        end
-
-        def occurrences
-          other_sexps.count
+        def other_sexps(members, i)
+          members.delete_at(i)
+          members.sort_by(&:file)
         end
       end
     end
--- a/spec/cc/engine/analyzers/javascript/main_spec.rb
+++ b/spec/cc/engine/analyzers/javascript/main_spec.rb
@@ -1,9 +1,8 @@
+require 'spec_helper'
 require 'cc/engine/analyzers/javascript/main'
 require 'cc/engine/analyzers/reporter'
 require 'cc/engine/analyzers/engine_config'
 require 'cc/engine/analyzers/file_list'
-require 'flay'
-require 'tmpdir'
 
 RSpec.describe CC::Engine::Analyzers::Javascript::Main, in_tmpdir: true do
   include AnalyzerSpecHelpers
@@ -16,7 +15,9 @@ RSpec.describe CC::Engine::Analyzers::Javascript::Main, in_tmpdir: true do
           console.log("hello JS!");
       EOJS
 
-      result = run_engine(engine_conf).strip
+      issues = run_engine(engine_conf).strip.split("\0")
+      result = issues.first.strip
+
       json = JSON.parse(result)
 
       expect(json["type"]).to eq("issue")
@@ -27,13 +28,13 @@ RSpec.describe CC::Engine::Analyzers::Javascript::Main, in_tmpdir: true do
         "path" => "foo.js",
         "lines" => { "begin" => 1, "end" => 1 },
       })
-      expect(json["remediation_points"]).to eq(297000)
+      expect(json["remediation_points"]).to eq(33_000)
       expect(json["other_locations"]).to eq([
         {"path" => "foo.js", "lines" => { "begin" => 2, "end" => 2} },
         {"path" => "foo.js", "lines" => { "begin" => 3, "end" => 3} }
       ])
-      expect(json["content"]["body"]).to match /This issue has a mass of `99`/
-      expect(json["fingerprint"]).to eq("55ae5d0990647ef496e9e0d315f9727d")
+      expect(json["content"]["body"]).to match /This issue has a mass of `11`/
+      expect(json["fingerprint"]).to eq("c4d29200c20d02297c6f550ad2c87c15")
     end
 
     it "prints an issue for similar code" do
@@ -43,7 +44,9 @@ RSpec.describe CC::Engine::Analyzers::Javascript::Main, in_tmpdir: true do
           console.log("helllllllllllllllllo JS!");
       EOJS
 
-      result = run_engine(engine_conf).strip
+      issues = run_engine(engine_conf).strip.split("\0")
+      result = issues.first.strip
+
       json = JSON.parse(result)
 
       expect(json["type"]).to eq("issue")
@@ -54,13 +57,13 @@ RSpec.describe CC::Engine::Analyzers::Javascript::Main, in_tmpdir: true do
         "path" => "foo.js",
         "lines" => { "begin" => 1, "end" => 1 },
       })
-      expect(json["remediation_points"]).to eq(99000)
+      expect(json["remediation_points"]).to eq(33_000)
       expect(json["other_locations"]).to eq([
         {"path" => "foo.js", "lines" => { "begin" => 2, "end" => 2} },
         {"path" => "foo.js", "lines" => { "begin" => 3, "end" => 3} }
       ])
-      expect(json["content"]["body"]).to match /This issue has a mass of `33`/
-      expect(json["fingerprint"]).to eq("55ae5d0990647ef496e9e0d315f9727d")
+      expect(json["content"]["body"]).to match /This issue has a mass of `11`/
+      expect(json["fingerprint"]).to eq("d9dab8e4607e2a74da3b9eefb49eacec")
     end
 
     it "skips unparsable files" do
@@ -91,8 +94,8 @@ RSpec.describe CC::Engine::Analyzers::Javascript::Main, in_tmpdir: true do
           <a className='button button-primary full' href='#' onClick={this.onSubmit.bind(this)}>Login</a>
     EOJSX
 
-    result = run_engine(engine_conf).strip
-    issues = result.split("\0")
+    issues = run_engine(engine_conf).strip.split("\0")
+
     expect(issues.length).to eq 1
   end
 
--- a/spec/cc/engine/analyzers/php/main_spec.rb
+++ b/spec/cc/engine/analyzers/php/main_spec.rb
@@ -1,9 +1,8 @@
+require 'spec_helper'
 require 'cc/engine/analyzers/php/main'
 require 'cc/engine/analyzers/reporter'
 require 'cc/engine/analyzers/engine_config'
 require 'cc/engine/analyzers/file_list'
-require 'flay'
-require 'tmpdir'
 
 RSpec.describe CC::Engine::Analyzers::Php::Main, in_tmpdir: true do
   include AnalyzerSpecHelpers
@@ -29,7 +28,9 @@ RSpec.describe CC::Engine::Analyzers::Php::Main, in_tmpdir: true do
           }
       EOPHP
 
-      result = run_engine(engine_conf).strip
+      issues = run_engine(engine_conf).strip.split("\0")
+      result = issues.first.strip
+
       json = JSON.parse(result)
 
       expect(json["type"]).to eq("issue")
@@ -40,17 +41,18 @@ RSpec.describe CC::Engine::Analyzers::Php::Main, in_tmpdir: true do
         "path" => "foo.php",
         "lines" => { "begin" => 2, "end" => 6 },
       })
-      expect(json["remediation_points"]).to eq(176000)
+      expect(json["remediation_points"]).to eq(44_000)
       expect(json["other_locations"]).to eq([
         {"path" => "foo.php", "lines" => { "begin" => 10, "end" => 14} },
       ])
-      expect(json["content"]["body"]).to match /This issue has a mass of `44`/
-      expect(json["fingerprint"]).to eq("667da0e2bab866aa2fe9d014a65d57d9")
+      expect(json["content"]["body"]).to match /This issue has a mass of `11`/
+      expect(json["fingerprint"]).to eq("8234e10d96fd6ef608085c22c91c9ab1")
     end
 
     it "runs against complex files" do
       FileUtils.cp(fixture_path("symfony_configuration.php"), File.join(@code, "configuration.php"))
-      result = run_engine(engine_conf).strip
+      issues = run_engine(engine_conf).strip.split("\0")
+      result = issues.first.strip
 
       expect(result).to match "\"type\":\"issue\""
     end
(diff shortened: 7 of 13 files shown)


lib/cc/engine/analyzers/python/main.rb
```diff
@@ -12,11 +12,20 @@ module Python
...
           DEFAULT_PATHS = ["**/*.py"]
-          DEFAULT_MASS_THRESHOLD = 40
-          BASE_POINTS = 1000
+          DEFAULT_MASS_THRESHOLD = 32
+          BASE_POINTS = 1_500_000
+          POINTS_PER_OVERAGE = 30_000
+
+          def calculate_points(issue)
+            BASE_POINTS + (overage(issue) * POINTS_PER_OVERAGE)
+          end
```
ABaldwinHunter: Note: I thought grades would get worse with this change, but overall they seem to be about the same. 

https://gist.github.com/ABaldwinHunter/e280310b4987766efd8c

ABaldwinHunter: Updating these stats

spec/cc/engine/analyzers/python/main_spec.rb
```diff
@@ -27,7 +27,7 @@
         "path" => "foo.py",
         "lines" => { "begin" => 1, "end" => 1 },
       })
-      expect(json["remediation_points"]).to eq(54000)
+      expect(json["remediation_points"]).to eq(3000000)
```
wfleming: :lipstick: `3_000_000`

spec/cc/engine/analyzers/python/main_spec.rb
```diff
@@ -54,7 +54,7 @@
         "path" => "foo.py",
         "lines" => { "begin" => 1, "end" => 1 },
       })
-      expect(json["remediation_points"]).to eq(18000)
+      expect(json["remediation_points"]).to eq(1920000)
```
wfleming: :lipstick: `1_920_000`

lib/cc/engine/analyzers/python/main.rb
wfleming: Re the comparisons of grades between classic & platform: it looks like these are overall grades across all categories? If we're trying to tune the engines, I think we should be trying to compare this engine specifically to it's classic equivalent (duplication category smells). Otherwise I think we're comparing at too high a level to have confidence in a given engine's tuning. This is related to what was discussed in IPM today. I'd love to see the sum of remediation points for duplication issues between classic & platform.

ABaldwinHunter: @wfleming I think that's a good point, and so the table actually includes a column comparing only duplication and complexity between Platform and Classic, and omitting style.

The complexity issues have already been tuned, so I expect them to have roughly the same impact. It could be interesting to see precisely how many points per category each repo gets, but I'm not sure it's that much more helpful than comparing issues and overall GPA?

If we think that'd be useful to see though, I can work on some reverse engineering of remediation points and add a table of those. 

After updating the stats, it looks like the grades are significantly worse but no incredibly dissimilar from Classic.

spec/cc/engine/analyzers/python/main_spec.rb
ABaldwinHunter: :+1:

spec/cc/engine/analyzers/python/main_spec.rb
ABaldwinHunter: :+1:

lib/cc/engine/analyzers/python/main.rb
wfleming: >  it looks like the grades are significantly worse but no incredibly dissimilar from Classic.

…What? I'm not sure what you mean here. "Significantly worse" & "not incredibly dissimilar" sound contradictory.

> If we think that'd be useful to see though, I can work on some reverse engineering of remediation points and add a table of those.

I don't think you need to reverse engineer anything: you can query for duplication issues in the `smells` mongo table for the respective snapshots you're looking at, and compare the remediation points for the smell records that seem to correlate (same file/location) between classic & platform. The number of duplication issues I saw in that gist seemed low enough that correlating by hand doesn't seem arduous.

There is a direct relationship between remediation points & repo grade. So if we want to feel confident that duplication is going to result in similar grades on platform as it did on classic, I think the best way to know that is by comparing the solid numbers we get from duplication issues between the two architectures, without all the other categories potentially muddying the waters.

ABaldwinHunter: @wfleming 
> it looks like the grades are significantly worse but no incredibly dissimilar from Classic.

 it looks like the grades are significantly worse than currently on Platform, but not dissimilar from Classic.

> There is a direct relationship between remediation points & repo grade.

True. It's important to note here though that we're not only comparing total remediation points, but also the variety of duplication issues found. 

I can certainly query the database and make some stats if we're interested in seeing how many duplication-based remediation points are reported for a given repo or given issue on Classic and Platform. 

My gut from the work so far is that the complexity-with-duplication grades are reflective of the duplication point differences.

wfleming: > it looks like the grades are significantly worse than currently on Platform, but not dissimilar from Classic.

Ah, thanks, that makes sense.

ABaldwinHunter:
@codeclimate/review @wfleming 

I made a script to collect stats that compare python duplication remediation points on classic, current platform, and platform with proposed updates: https://gist.github.com/ABaldwinHunter/da407f6657ef30a05069

The proposed new points formula brings our analysis closer to Classic, but in some cases assigns significantly steeper penalties than on Classic for the same issue. I think classic may add additional remediation cost per occurrence when multiple instances of code are found (they show up as individual smells, but share a fingerprint) - but am currently double-checking.

To reduce the impact of threshold differences, I might consider reducing the `per_cost` variable further (already reduced from 50_000 to 30_000).

I've also only run these stats on Flask. I could run them on a few additional repositories.

wfleming:
Nice to see that new points values are the same number of significant digits as classic! The difference between new platform/classic looks to be a ratio of ~2.17 on average. The remediation points -> GPA algorithm is an exponential step-off (i.e. twice as many points means one letter grade drop), so in terms of resultant grades at least, it seems like more `per_cost` tweaking might be valuable.

ABaldwinHunter:
@codeclimate-review 

Ready for re-review

Summary:

- Update remediation points calculation (matches Classic, with one difference: don't penalize extra for identical code)
- Bug fix: Report issue for each file with duplication, instead of only first location
- Use mass as node size, not Flay score

ABaldwinHunter:
cc @brynary

ABaldwinHunter:
Note: this PR presents a few changes. I like keeping them together because they're finished and feel related, but can see the argument for splicing PR if that'd be cleaner.

config/contents/duplicated_code.md.erb
```diff
@@ -9,7 +9,7 @@ When you violate DRY, bugs and maintenance problems are sure to follow. Duplicat
 ## Issue Mass
 
 Duplicated code has a calculated mass, which can be thought of as a measure of how much logic has been duplicated.
-This issue has a mass of `<%= issue.mass %>`: if you would like to change the minimum mass that will be reported as an issue, please see the details in [`codeclimate-duplication`'s documentation](https://github.com/codeclimate/codeclimate-duplication).
+This issue has a mass of `<%= mass %>`: if you would like to change the minimum mass that will be reported as an issue, please see the details in [`codeclimate-duplication`'s documentation](https://github.com/codeclimate/codeclimate-duplication).
```
wfleming: If we're putting this in the description, I think we can drop it from the read up. I thought that got done?

lib/cc/engine/analyzers/python/main.rb
```diff
@@ -12,11 +12,20 @@ module Python
         class Main < CC::Engine::Analyzers::Base
           LANGUAGE = "python"
           DEFAULT_PATHS = ["**/*.py"]
-          DEFAULT_MASS_THRESHOLD = 40
-          BASE_POINTS = 1000
+          DEFAULT_MASS_THRESHOLD = 32
+          BASE_POINTS = 1_500_000
+          POINTS_PER_OVERAGE = 50_000
+
+          def calculate_points(mass)
```
wfleming: This equation should be the same for all languages, shouldn't it (modulo different constants which are, well, constants). Can we lift this to a base class?

lib/cc/engine/analyzers/python/main.rb
```diff
@@ -12,11 +12,20 @@ module Python
...
+            BASE_POINTS + (overage(mass) * POINTS_PER_OVERAGE)
+          end
 
           private
 
+          def overage(mass)
```
wfleming: Ditto: since this method should be the same for all languages it's be nice to see it extracted.

config/contents/duplicated_code.md.erb
wfleming: Oops, I realized #77 hasn't shipped yet.

lib/cc/engine/analyzers/violation.rb
```diff
@@ -45,7 +64,9 @@ def sorted_hashes
         end
 
         def other_sexps
-          @other_locations ||= sorted_hashes.drop(1)
+          sorted = sorted_hashes
```
wfleming: Doing a mutation on the array here is unfortunate. Can we do better? In fact, tracking the active hash by index up in `#initialize` feels like it could be cleaner.

We could track the actual current directly instead:

```
def initialize((language_strategy, issue, hashes, current_sexp = nil)
    ...
    @location = current_sexp || sorted_hashes.first
end

def other_sexps
  sorted_hashes.reject { |other| other == current_sexp }
end
```

lib/cc/engine/analyzers/violation.rb
```diff
@@ -32,12 +39,24 @@ def report_name
           "#{current_sexp.file}-#{current_sexp.line}"
         end
 
+        def mass # ABH Reverse engineer the actual mass from flay score.
```
wfleming: This feels like the big change here, and it's applicable across the board. If this can be cleanly done in it's own PR, I'd say go for that: it not, I would suggest editing the PR title/description to mention this in more detail, since those things will end up in commits & be the likely way people will search for the PR for historical context in the future. Your call, though.

lib/cc/engine/analyzers/reporter.rb
```diff
@@ -38,9 +38,11 @@ def report
           flay.report(StringIO.new).each do |issue|
             violation = new_violation(issue)
 
-            unless reports.include?(violation.report_name)
-              reports.add(violation.report_name)
-              io.puts "#{violation.format.to_json}\0"
+            violation.occurrences.each do |occurrence|
```
wfleming: This is the other big change, to me. I have the same thinking about this as about the mass calculation change.

lib/cc/engine/analyzers/violation.rb
```diff
@@ -32,12 +39,24 @@ def report_name
...
+          if issue.identical?
+            flay_score / (occurrences.length**2)
```
wfleming: :lipstick: I'm surprised rubocop didn't complain about lack of space around the `**` operator.

lib/cc/engine/analyzers/violation.rb
```diff
@@ -91,19 +112,15 @@ def fingerprint
           digest << "-"
           digest << current_sexp.mass.to_s
           digest << "-"
-          digest << occurrences.to_s
+          digest << occurrences.length.to_s
```
wfleming: With the change to reporting separate issues for each occurrence, this leads to some interesting behavior. Let's say I have identical code in 3 files & I copy-paste it to a fourth. The result will be that the previous 3 issues will be "fixed" and I'll have 4 "new" issues. If code in file A hasn't changed then it feels odd that the file A specific issue for the duplication got marked fixed and then new.

I'm not suggesting this should change, just thinking aloud abut whether this is the optimal choice on our part. Since all of the other locations are attached to each instance of the issue, it might be ok to not use occurrences as part of the fingerprint.

spec/cc/engine/analyzers/javascript/main_spec.rb
```diff
@@ -27,13 +29,13 @@
         "path" => "foo.js",
         "lines" => { "begin" => 1, "end" => 1 },
       })
-      expect(json["remediation_points"]).to eq(297000)
+      expect(json["remediation_points"]).to eq(33000)
```
wfleming: :lipstick: `33_000`

spec/cc/engine/analyzers/javascript/main_spec.rb
```diff
@@ -54,13 +58,13 @@
         "path" => "foo.js",
         "lines" => { "begin" => 1, "end" => 1 },
       })
-      expect(json["remediation_points"]).to eq(99000)
+      expect(json["remediation_points"]).to eq(33000)
```
wfleming: :lipstick: `33_000`

spec/cc/engine/analyzers/javascript/main_spec.rb
```diff
@@ -91,8 +95,8 @@
           <a className='button button-primary full' href='#' onClick={this.onSubmit.bind(this)}>Login</a>
     EOJSX
 
-    result = run_engine(engine_conf).strip
-    issues = result.split("\0")
+    issues = run_engine(engine_conf).strip.split("\0")
```
wfleming: Nice to see these little snippets getting compressed!

lib/cc/engine/analyzers/violation.rb
wfleming: Can we just use the `#mass` method on the `Sexp`? This code looks like it's trying to reverse engineer flay's internal score calculation from `Sexp#mass`, but we have the `Sexp` available here.

lib/cc/engine/analyzers/violation.rb
ABaldwinHunter: Good point. The behavior in Classic is that all four issues have the same fingerprint (which actually, also seems a little problematic, depending how we using the fingerprinting). 

Maybe we could just nuke line 115?

spec/cc/engine/analyzers/javascript/main_spec.rb
ABaldwinHunter: :+1:

lib/cc/engine/analyzers/violation.rb
ABaldwinHunter: Okay, I can see if the Sexp has the mass/node_size attached. If so, that would be great.

lib/cc/engine/analyzers/violation.rb
wfleming: I'm :+1:  on nuking it, personally. Might be worth getting more opinions.

pbrisbin: >  The behavior in Classic is that all four issues have the same fingerprint 

In general, if Classic behaves a certain way and we don't understand it well or have a definitive reason to believe it's buggy or inapplicable to Platform, we should err on the side of preserving it -- IMHO.

My memory is hazy, but I think the logic in the UI that collapses multiple instances of duplication within a single file to just one rendered issue (with other locations attached) relies on the fact that all the instances are the same fingerprint.

It also sounds like giving them all the same fingerprint means that, in your example @wfleming, it would not show up as a new issue since there's no new fingerprints in the *To* snapshot. That might actually make sense: it's not a new issue, you've just made an existing issue worse. I think that might even render as such since to remediation points would increase.

wfleming: Ok, thanks, @pbrisbin. That makes sense as a description of desired behavior. If that's the behavior we want, I think we have to rethink the fingerprinting here more -- right now fingerprints are different for each instance because the hash uses `current_sexp.file`.

Making it so that a new instance of the same duplication keeps the old fingerprint is an interesting wrinkle. Do you know how we did that in classic? I can't think of a solution that would be 100% reliable.

pbrisbin: > Do you know how we did that in classic?

I don't. And please don't take my comments as definitely accurate.

This is a thorny problem with lots of landmines, and the desired behavior is pretty subtle (for example, is cross-file handling different than same-file?). I think we might need to pick @brynary 's brain about desired behavior, then dig into how the Classic code is implemented.

lib/cc/engine/analyzers/python/main.rb
ABaldwinHunter: I like lifting to a base class as well! However, since we're updating the tuning algorithm language-by-language, I was keeping separate. 

Some of the constants might be different per language as well.

I lean toward favoring moving into a base class as a nice separate refactor PR after, once changes are in place.

ABaldwinHunter: WDYT?

lib/cc/engine/analyzers/violation.rb

lib/cc/engine/analyzers/violation.rb
ABaldwinHunter: Also good point about preserving changes for future context.

lib/cc/engine/analyzers/violation.rb
ABaldwinHunter: > My memory is hazy, but I think the logic in the UI that collapses multiple instances of duplication within a single file to just one rendered issue (with other locations attached) relies on the fact that all the instances are the same fingerprint.

@pbrisbin The UI doesn't do this - at least for the `issues > duplication` tab.

It shows all of the issues. 

e.g. https://codeclimate.com/repos/568aeda48c5d387d9200048d/issues/categories/duplication

in `flask/blueprints.py`

pbrisbin: I would've guessed this logic is in the compare view(s), but I can't find it. The only `uniq`-ing by fingerprint seems to be in `SmellsCounter`:

https://github.com/codeclimate/app/blob/master/app/models/smells_counter.rb#L50

So perhaps that's the only place we want special behavior: in the category counts.

lib/cc/engine/analyzers/python/main.rb
wfleming: I fully expect constants to continue to differ between language. I'm fine with doing it as a refactor pass later, sure.

lib/cc/engine/analyzers/violation.rb
wfleming: This brings up an interesting point about the remedation points algorithm as well: per-instance issues means that a new instance of the same duplication ends up creating a new instance on our end, but does make existing issues worse.

ABaldwinHunter: Thanks for your guys' input!

Sounds like there's a lot of philosophy to consider.

I think those things are worth investigating and discussing, but we could also make some of the small changes at the same time and incrementally move our sculpture towards the desired final shape - releasing the figure from within the marble block. (More philosophy ;) )

E.g. ship tuning points algorithm update and reporting issue for each duplication instance. 
Maybe thinking about an occurrence of duplication as many issues instead of one is fine?

Going to extract some of the tiny cuts into separate PRs.

brynary:
I have not reviewed this PR or read the context. I just saw one comment @pbrisbin made and it set off a flag...

The UI does _not_ and should not attempt to "collapse" multiple Issues of duplication within the same file caused by multiple occurrences of the same code structure within a single file.

If a duplication is within a single file, and there are e.g. 2 occurrences, we should produce and render 2 Issues on the page. The count should show "2" in the sidebar also (not 1).

This is an important change we made over a year ago and is important. :) /c @pbrisbin @ABaldwinHunter

pbrisbin:
@brynary Perfect, thanks for weighing in. It seems I'm remember when such a feature was discussed or implemented but wasn't around, or don't remember, when it was reverted or rejected. It's certainly easier to not have that UI logic, so that's good.

Can you speak generally to duplication fingerprinting? Does each instance get a unique fingerprint (it sounds like yes). Would you consider https://github.com/codeclimate/app/blob/master/app/models/smells_counter.rb#L50 a bug / hold-over from when some sort of duplication collapsing was present in the UI? It also sounds like this is a yes, but the answer could depend on if there are any other cases where duplications might be, well, duplicated and need `uniq`ing.

ABaldwinHunter:
@pbrisbin I'm also leaning toward us saying that each duplication instance gets a unique fingerprint, and then any new instances or removals appear as `new` and `fixed` issues, rather than one `improved` situation.

But curious what B thinks.

pbrisbin:
Seems pretty well answered to me:

> If a duplication is within a single file, and there are e.g. 2 occurrences, we should produce and render 2 Issues on the page. The count should show "2" in the sidebar also (not 1).

This discussion is also not meant to decide on a new Right Thing -- we're trying to get to the bottom of how Classic works so we can preserve it.

lib/cc/engine/analyzers/violation.rb
ABaldwinHunter: @wfleming Actually, that implementation doesn't work. The `sexp` doesn't have a location attached to it, so multiple sexps could be equal.

wfleming: But... [we get the location *from* the sexp](https://github.com/codeclimate/codeclimate-duplication/blob/abh-python-tuning/lib/cc/engine/analyzers/violation.rb#L84-L86). Is it just not taken into account by `Sexp#==`?

brynary:
@pbrisbin I believe that on Classic each occurrence does get the same fingerprint, but that was mostly incidental relative to the context. I suspect either way would be acceptable for now, but probably safest to use what Classic does to avoid any surprises.

lib/cc/engine/analyzers/violation.rb
wfleming: I guess part of what feels very strange to me is that `Violation` has this internal logic for determining which member of a set is it's "primary" and which are the "other" instances. That feels more like it should be the responsibility of whatever calls `Violation.new`, and `Violation.new` should get e.g. `current_sexp` & `other_sexps`.

ABaldwinHunter: I definitely agree that the class seems wonky. Just wasn't sure if fixing it was in the scope of this PR. 

Added an `issue` class - corresponding to occurrence of duplication code - to address this problem.

ABaldwinHunter:
@wfleming Ready for re-review!

I think I addressed the code concerns you mentioned - 
1. reorganize violation handling of current_ and other_ sexps using an `Issue` class
2. use `sexp.mass` instead of flay score reverse engineering
3. don't use other_sexps.count in fingerprint to avoid false positives

Thing I didn't do and am trying to squeeze by: break into separate PRs and commits.

Included clear bulleted summary in the commit. I don't think at this point reverse engineering the commit history is worth while, especially because the changes are intertwine, but can go ahead and take the time to do it if we prefer it? 

cc @codeclimate/review @brynary

lib/cc/engine/analyzers/issue.rb
```diff
@@ -0,0 +1,89 @@
...
+            "description": description,
+            "categories": ["Duplication"],
+            "location": format_location,
+            "remediation_points": calculate_points,
+            "other_locations": format_other_locations,
+            "content": content_body,
+            "fingerprint": fingerprint,
+          }
+        end # rubocop:enable Metrics/MethodLength
+
```
ABaldwinHunter: This code is largely a clean transplant from `Violation`. Added comments to disable method length check here bc it didn't seem useful.

lib/cc/engine/analyzers/issue.rb
```diff
@@ -0,0 +1,89 @@
...
+        def report_name
+          "#{current_sexp.file}-#{current_sexp.line}"
+        end
+
+        def mass
+          current_sexp.mass
+        end
```
ABaldwinHunter: This change is important: using the `mass` form the sexp gets us nodesize. Taking it from flay output in `issue.mass` got us flay score previously.

spec/cc/engine/analyzers/php/main_spec.rb
```diff
@@ -91,11 +93,6 @@
     end
   end
 
-  def printed_issue
-    issue = {"type":"issue","check_name":"Identical code","description":"Similar code found in 1 other location","categories":["Duplication"],"location":{"path":"foo.php","lines":{"begin":2,"end":6}},"remediation_points":176000,"other_locations":[{"path":"foo.php","lines":{"begin":10,"end":14}}],"content":{"body": read_up}}
-    issue.to_json + "\0\n"
-  end
-
```
ABaldwinHunter: removed unused method and refactored specs to use `spec_helper` at the top.

lib/cc/engine/analyzers/issue.rb
```diff
@@ -0,0 +1,89 @@
...
+          }
+        end
+
+        def content_body
+          { "body": ViolationReadUp.new(mass).contents }
+        end
+
+        def fingerprint
+          digest = Digest::MD5.new
+          digest << current_sexp.file
```
wfleming: I know this has been discussed to death here, so pardon me for bringing it up again, but I'm still not 100% clear on what we decided the right thing to do here is.

This implementation will result in the fingerprint of two duplication issues in the same file (even duplication of different pieces of code that happen to have the same mass) having the same fingerprint. You're saying that this collection of data (file name, mass, check name) was how classic calculated the fingerprint, and such a collision is expected?

Since our comparison algorithm relies on fingerprints, I'm pretty sure this could result in an analysis producing new duplication issues, but the PR would still show "no new issues" because of the fingerprint collision. That does not sound like desired behavior to me, so I'm still not convinced that "do what classic did" is the right call here.

lib/cc/engine/analyzers/violation.rb
```diff
@@ -1,52 +1,32 @@
...
-            "content": content_body,
-            "fingerprint": fingerprint
-          }
+        def occurrences
+          hashes.map.with_index do |sexp, i|
+            Issue.new(language_strategy: language_strategy, check_name: check_name, current_sexp: sexp, other_sexps: other_sexps(hashes.dup, i))
+          end
         end
 
         def report_name
```
wfleming: I don't think this method is actually needed on this class anymore? It's just relevant on `Issue` now, right?

lib/cc/engine/analyzers/violation.rb
```diff
@@ -1,52 +1,32 @@
...
           "#{current_sexp.file}-#{current_sexp.line}"
         end
 
-        private
-
-        attr_reader :language_strategy, :hashes
-
-        def current_sexp
-          @location ||= sorted_hashes.first
+        def mass
```
wfleming: I don't think this method is actually needed on this class anymore? It's just relevant on `Issue` now, right?

lib/cc/engine/analyzers/issue.rb
ABaldwinHunter: Ha, No worries. :)

This is the code I found in classic: https://github.com/codeclimate/analyzer/blame/master/lib/quality/smells/duplication.rb

It uses the constant name, so the matching fingerprints would only be for duplication issues within the same file. Looking at the commit history, it's not completely clear to me why that was done - it doesn't seem to have any impact now in the UI- and I agree that it's a bit nonsensical for reporting new issues 

`IssuesDelta` In Finalizer: https://github.com/codeclimate/finalizer/blob/master/lib/cc/finalizer/issues_delta.rb#L21

lib/cc/engine/analyzers/violation.rb
ABaldwinHunter: :+1

ABaldwinHunter: Good catch. Thanks

lib/cc/engine/analyzers/violation.rb
ABaldwinHunter: :+1:  right again, thanks

lib/cc/engine/analyzers/issue.rb
wfleming: We did a lot of work semi-recently to fix comparison differences between app/compare & finalizer. And when we shipped the new PR header, it was considered critical that the app/compare's pass/fail logic match finalizer's. I think this fingerprinting algorithm will break those expectations.

Example:

in the master snapshot, file `foo.rb` has the following contents:

```
class A
  def foo
     # some large duplicated code block here
  end

  def bar
     # the same code block here
  end
end
```

That will result in 2 issues with the same fingerprint, but it's the master snapshot, so who cares.

In the pull request snapshot, file `foo.rb` looks like this:

```
class A
  def foo
     # some large duplicated code block here
  end

  def bar
     # the same code block here
  end

  def baz
    # the same code block here, again
  end
end
```

Now we have 3 issues on `foo.rb`, all with the same snapshot. (And the newest issue for `baz` doesn't even have to be same code: as I said, it could happen to be totally different code duplicated elsewhere which happens to have the same mass as `foo` and `bar`.)

Finalizer is going to see no new fingerprints, and pass the PR. If the user goes to the PR compare page, I *think* they'll see the "passed" header, but in the issues area they'll see 1 new issue.

ABaldwinHunter:
Closing because this branch has been harvested and starfished into others.

PR #80, opened 2016-01-20
Install npm as part of docker build

ABaldwinHunter:
@codeclimate/review 

Fixes the following error I began seeing locally:

```
Removing intermediate container 85df8e8d7b2e
Step 14 : RUN npm install
 ---> Running in 1b4d48c42a57
/bin/sh: 1: npm: not found
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,7 @@ COPY vendor/php-parser/composer.json /usr/src/app/vendor/php-parser/
 COPY vendor/php-parser/composer.lock /usr/src/app/vendor/php-parser/
 
 RUN curl --silent --location https://deb.nodesource.com/setup_5.x | bash -
-RUN apt-get update && apt-get install -y nodejs python openssh-client php5-cli php5-json
+RUN apt-get update && apt-get install -y nodejs npm python openssh-client php5-cli php5-json
 RUN gem install bundler --no-ri --no-rdoc && \
     bundle install -j 4 && \
     curl -sS https://getcomposer.org/installer | php
--- a/circle.yml
+++ b/circle.yml
@@ -11,7 +11,7 @@ machine:
 dependencies:
   override:
     - docker info
-    - docker build -t=$PRIVATE_REGISTRY/$CIRCLE_PROJECT_REPONAME:b$CIRCLE_BUILD_NUM .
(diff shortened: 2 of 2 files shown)


ABaldwinHunter:
Closing because of circle failures. Investigating why the docker build doesn't work locally without `npm`, and can't work on circle with it. :(

ABaldwinHunter:
Rebuilt the image locally without cache and that did the trick fwiw.

Dockerfile
```diff
@@ -9,7 +9,7 @@ COPY vendor/php-parser/composer.json /usr/src/app/vendor/php-parser/
 COPY vendor/php-parser/composer.lock /usr/src/app/vendor/php-parser/
 
 RUN curl --silent --location https://deb.nodesource.com/setup_5.x | bash -
-RUN apt-get update && apt-get install -y nodejs python openssh-client php5-cli php5-json
+RUN apt-get update && apt-get install -y nodejs npm python openssh-client php5-cli php5-json
```
pbrisbin: @ABaldwinHunter your issue is because of the dependency between the installation here and the nodesource install a line above. You should combine them to one `RUN` invocation with `&&` so that they are cached (and invalidated) together.

circle.yml
```diff
@@ -11,7 +11,7 @@ machine:
 dependencies:
   override:
     - docker info
-    - docker build -t=$PRIVATE_REGISTRY/$CIRCLE_PROJECT_REPONAME:b$CIRCLE_BUILD_NUM .
+    - docker build --rm -t=$PRIVATE_REGISTRY/$CIRCLE_PROJECT_REPONAME:b$CIRCLE_BUILD_NUM .
```
pbrisbin: FYI: `--rm` doesn't work on Circle because we don't have permissions to remove layers, it'll just spit warnings to stderr. Also, since 1.9, `--rm` is the default.

Dockerfile
ABaldwinHunter: @pbrisbin cool, thanks for the explanation!

circle.yml
ABaldwinHunter: GTK Thanks!

Dockerfile
ABaldwinHunter: Added in pull: https://github.com/codeclimate/codeclimate-duplication/compare/abh-dockerfile-update?expand=1

PR #81, opened 2016-01-20
Include erb files for analysis, following Flay

ABaldwinHunter:
This erb-processing code is lifted directly from flay:
https://github.com/seattlerb/flay/blob/master/lib/flay_erb.rb#L13

@codeclimate/review @noahd1

### DIFF
--- a/lib/cc/engine/analyzers/ruby/main.rb
+++ b/lib/cc/engine/analyzers/ruby/main.rb
@@ -1,3 +1,4 @@
+require "erubis"
 require "flay"
 require "json"
 require "cc/engine/analyzers/reporter"
@@ -10,12 +11,12 @@ module CC
         class Main < CC::Engine::Analyzers::Base
           LANGUAGE = "ruby"
           DEFAULT_PATHS = [
+            "**/*.erb",
             "**/*.rb",
             "**/*.rake",
             "**/Rakefile",
             "**/Gemfile",
             "**/*.gemspec"
-
           ]
           DEFAULT_MASS_THRESHOLD = 18
           BASE_POINTS = 1_500_000
@@ -33,7 +34,37 @@ module CC
           end
 
           def process_file(file)
-            RubyParser.new.process(File.binread(file), file, TIMEOUT)
+            if File.extname(file) == ".erb"
+              process_erb file
+            else
+              RubyParser.new.process(File.binread(file), file, TIMEOUT)
+            end
+          end
+
+          def process_erb(file)
+            erb = File.binread(file)
+            ruby = Erubis.new(erb).src
+            RubyParser.new.process(ruby, file)
+          end
+
+          class Erubis < ::Erubis::Eruby
+            BLOCK_EXPR = /\s+(do|\{)(\s*\|[^|]*\|)?\s*\Z/
+
+            def add_expr_literal(src, code)
+              if code =~ BLOCK_EXPR
+                src << "@output_buffer.append= " << code
+              else
+                src << "@output_buffer.append=(" << code << ");"
+              end
+            end
+
+            def add_expr_escaped(src, code)
+              if code =~ BLOCK_EXPR
+                src << "@output_buffer.safe_append= " << code
+              else
+                src << "@output_buffer.safe_append=(" << code << ");"
+              end
+            end
           end
         end
       end
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -8,6 +8,8 @@ engines:
     enabled: true
   duplication:
     enabled: true
+    exclude_fingerprints:
+    - b2dc8dbd27916d8321e298f59fc2f431 # Erubis < ::Erubis::Eruby
     config:
       languages:
       - ruby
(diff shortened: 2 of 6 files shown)


lib/cc/engine/analyzers/ruby/main.rb
```diff
@@ -33,8 +34,45 @@ def overage(issue)
...
+            ruby = Erubis.new(erb).src
+            begin
+              RubyParser.new.process(ruby, file)
+            rescue => e
+              warn ruby if option[:verbose]
+              raise e
+            end
+          end
+
+          class Erubis < ::Erubis::Eruby # :nodoc:
```
wfleming: :lipstick: The `nodoc` comment is extraneous since don't use rdoc.

ABaldwinHunter: :+1:  thanks

lib/cc/engine/analyzers/ruby/main.rb
```diff
@@ -33,8 +34,47 @@ def overage(issue)
           end
 
           def process_file(file)
-            RubyParser.new.process(File.binread(file), file, TIMEOUT)
+            if File.extname(file) == ".erb"
```
wfleming: WDYT think about extracting all of this into a `Parser` class like the other languages have? Keeping that logic here made sense when it was one-liner, but I think this is complex enough that extracting makes sense.

ABaldwinHunter: I actually prefer the single file because I think the `main` class is fairly small (< 5 methods), and adding encapsulating parser files might be adding needless encapsulation.

Let me know if you feel strongly though - and I can make a change.

ABaldwinHunter:
For the CC duplication issue, I considered  extracting part of the method, but I think it adds more lines and complexity in this case. More readable as is.

ABaldwinHunter:
ping @wfleming @codeclimate/review Ready for re-review!

lib/cc/engine/analyzers/ruby/main.rb
```diff
@@ -33,7 +34,45 @@ def overage(issue)
...
+              process_erb file
+            else
+              RubyParser.new.process(File.binread(file), file, TIMEOUT)
+            end
+          end
+
+          # taken from Flay, https://github.com/seattlerb/flay/blob/master/lib/flay_erb.rb
+
```
wfleming: :lipstick: extra space

lib/cc/engine/analyzers/ruby/main.rb
```diff
@@ -33,7 +34,45 @@ def overage(issue)
...
+          def process_erb(file)
+            erb = File.read(file)
+
+            ruby = Erubis.new(erb).src
+            begin
+              RubyParser.new.process(ruby, file)
+            rescue => e
+              $stderr.puts ruby
```
wfleming: This seems a bit excessive, and could be very large.

We catch exceptions up the stack, and we're raising right below here anyway: we can probably lose the begin/rescue here. We'll get the filename up the stack & us/customers can user that for debugging.

wfleming:
Noticed 2 more small things. If we're going to exclude the duplication issue, maybe we should add it to the ignored fingerprints in config?

Otherwise, LGTM.

lib/cc/engine/analyzers/ruby/main.rb
ABaldwinHunter: :+1:  sounds good

lib/cc/engine/analyzers/ruby/main.rb
ABaldwinHunter: :+1:

ABaldwinHunter:
@codeclimate/review 

Ready for re-review!

@wfleming I 

- removed the begin/rescue clause and output to stderr [here](https://github.com/codeclimate/codeclimate-duplication/pull/81/files#diff-415d8ecf3775cdd67402dc77177f7b15R44), as suggested
- added codeclimate issue to finger_print excludes

lib/cc/engine/analyzers/ruby/main.rb
```diff
@@ -33,7 +34,37 @@ def overage(mass)
...
-            RubyParser.new.process(File.binread(file), file, TIMEOUT)
+            if File.extname(file) == ".erb"
+              process_erb file
+            else
+              RubyParser.new.process(File.binread(file), file, TIMEOUT)
+            end
+          end
+
+          def process_erb(file)
+            erb = File.read(file)
```
wfleming: This is a minor pondering that just occurred to me: we use `binread` above but just use `read` here. I suspect this is down to file encodings. You comfortable having these be different?

wfleming:
One small question, but LGTM.

lib/cc/engine/analyzers/ruby/main.rb
ABaldwinHunter: Thanks for the catch, `binread` sounds good to me.

I pulled the code from `flay` and tested it out locally without an issue, and didn't realize the intention behind using `File.binread`. Making switch

PR #82, opened 2016-01-20
dockerfile update

ABaldwinHunter:
Combine node install commands into single RUN

Avoids problematic behavior that occurs when the nodesource curl and
nodejs install commands are cached separately.

Following up on suggestion from https://github.com/codeclimate/codeclimate-duplication/pull/80/files#r50307883

@codeclimate/review


wfleming:
:+1:

PR #83, opened 2016-01-25
Report issue for each occurrence of duplication

ABaldwinHunter:
This change fixes a bug in which we reported only the first occurrence
(or sexp) of duplicated code in a set, and dropped the rest. The others
were included in other_locations but not reported as an independent issue.

@codeclimate/review 
cc @pbrisbin 

First step in breaking apart https://github.com/codeclimate/codeclimate-duplication/pull/78

### DIFF
--- a/lib/cc/engine/analyzers/violation.rb
+++ b/lib/cc/engine/analyzers/violation.rb
@@ -8,10 +8,11 @@ module CC
       class Violation
         attr_reader :issue
 
-        def initialize(language_strategy, issue, hashes)
+        def initialize(language_strategy:, issue:, current_sexp:, other_sexps:)
           @language_strategy = language_strategy
           @issue = issue
-          @hashes = hashes
+          @current_sexp = current_sexp
+          @other_sexps = other_sexps
         end
 
         def format
@@ -34,19 +35,7 @@ module CC
 
         private
 
-        attr_reader :language_strategy, :hashes
-
-        def current_sexp
-          @location ||= sorted_hashes.first
-        end
-
-        def sorted_hashes
-          @_sorted_hashes ||= hashes.sort_by(&:file)
-        end
-
-        def other_sexps
-          @other_locations ||= sorted_hashes.drop(1)
-        end
+        attr_reader :language_strategy, :other_sexps, :current_sexp
 
         def check_name
           if issue.identical?
--- a/spec/cc/engine/analyzers/javascript/main_spec.rb
+++ b/spec/cc/engine/analyzers/javascript/main_spec.rb
@@ -16,7 +16,8 @@ RSpec.describe CC::Engine::Analyzers::Javascript::Main, in_tmpdir: true do
           console.log("hello JS!");
       EOJS
 
-      result = run_engine(engine_conf).strip
+      issues = run_engine(engine_conf).strip.split("\0")
+      result = issues.first.strip
       json = JSON.parse(result)
 
       expect(json["type"]).to eq("issue")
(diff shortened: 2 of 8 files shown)


lib/cc/engine/analyzers/violation.rb
```diff
@@ -96,8 +85,8 @@ def fingerprint
         end
 
         def description
-          description = "#{check_name} found in #{occurrences} other location"
-          description += "s" if occurrences > 1
+          description = "#{check_name} found in #{(other_sexps.length)} other location"
+          description += "s" if other_sexps.length > 1
```
pbrisbin: Now that we had to bring back `#occurrences`, I think you can revert this too.

spec/cc/engine/analyzers/javascript/main_spec.rb
```diff
@@ -91,8 +95,8 @@
           <a className='button button-primary full' href='#' onClick={this.onSubmit.bind(this)}>Login</a>
     EOJSX
 
-    result = run_engine(engine_conf).strip
-    issues = result.split("\0")
+    issues = run_engine(engine_conf).strip.split("\0")
```
pbrisbin: I'd back this out. It's not a *terrible* case of drive-by, but just to flex the "essential changes only" muscles.

spec/cc/engine/analyzers/ruby/main_spec.rb
```diff
@@ -43,10 +45,43 @@ module CC::Engine::Analyzers
         expect(json["other_locations"]).to eq([
           {"path" => "foo.rb", "lines" => { "begin" => 9, "end" => 13} },
         ])
+
```
pbrisbin: Accidental whitespace.

spec/cc/engine/analyzers/ruby/main_spec.rb
```diff
@@ -43,10 +45,43 @@ module CC::Engine::Analyzers
...
+              it 'returns true' do
+                expect(subject.js?).to be true
+              end
+            end
+        EORUBY
+
+        issues = run_engine(engine_conf).strip.split("\0")
+
+        expect(issues.length).to eq(3)
```
pbrisbin: I think we need to get some better coverage, and can probably do so with a spec on `Violations`. Here you assert we get three `Violation`s out, which is good -- but I'd like to also confirm some attributes on each violation, that the main and other locations are accurate mostly.

lib/cc/engine/analyzers/violation.rb

spec/cc/engine/analyzers/ruby/main_spec.rb
ABaldwinHunter: :+1:

spec/cc/engine/analyzers/ruby/main_spec.rb
ABaldwinHunter: Sounds good. Adding a spec

ABaldwinHunter:
@pbrisbin Ready for re-review!

ABaldwinHunter:
Removed any drive-by changes and added spec.

spec/cc/engine/analyzers/javascript/main_spec.rb
```diff
@@ -93,6 +95,7 @@
 
     result = run_engine(engine_conf).strip
     issues = result.split("\0")
+
```
pbrisbin: Accidental change.

pbrisbin:
This LGTM.

I fully expect this to have a (potentially strong) impact on grades. We're taking what was a single thing representing N instances of duplication with X remediation_points and now reporting N things each with X remediation_points. This means the total remediation_points for the duplication will go from X to X*N. I'm fine with this, we'll discuss this and tune points and grades separately via a new PR, but it should be explained in the ChangeLog.

ABaldwinHunter:
:+1:  

The mass and point algorithm changes coming on the heels of this PR will also impact grades. The overall impact may be smaller grade impact on more files.

PR #84, opened 2016-01-26
Use sexp node size for issue mass, not flay score

ABaldwinHunter:
This change fixes a bug, and impacts both remediation points
calculation and the issue mass reported in UI.

Motivation: create parity with Classic, and present user more
meaningful mass number.

An important behavioral departure from Classic is that we are no longer penalizing
extra for identical (as opposed) to similar duplicated code: remediation points
reflect the amount of effort needed to repair an issue, and perfectly identical code
can actually be easier to fix than code that is similar.

For reference, flay score is:

Similar code
    node_size * number_of_occurrences_of_duplicated_issue

Identical code
    node_size * num_occurrences * num_occurrences

@codeclimate/review @pbrisbin 

Step two of breaking apart https://github.com/codeclimate/codeclimate-duplication/pull/78

### DIFF
--- a/config/contents/duplicated_code.md.erb
+++ b/config/contents/duplicated_code.md.erb
@@ -9,7 +9,7 @@ When you violate DRY, bugs and maintenance problems are sure to follow. Duplicat
 ## Issue Mass
 
 Duplicated code has a calculated mass, which can be thought of as a measure of how much logic has been duplicated.
-This issue has a mass of `<%= issue.mass %>`: if you would like to change the minimum mass that will be reported as an issue, please see the details in [`codeclimate-duplication`'s documentation](https://github.com/codeclimate/codeclimate-duplication).
+This issue has a mass of `<%= mass %>`: if you would like to change the minimum mass that will be reported as an issue, please see the details in [`codeclimate-duplication`'s documentation](https://github.com/codeclimate/codeclimate-duplication).
 
 ## Refactorings
 
--- a/lib/cc/engine/analyzers/analyzer_base.rb
+++ b/lib/cc/engine/analyzers/analyzer_base.rb
@@ -36,8 +36,8 @@ module CC
           engine_config.mass_threshold_for(self.class::LANGUAGE) || self.class::DEFAULT_MASS_THRESHOLD
         end
 
-        def calculate_points(issue)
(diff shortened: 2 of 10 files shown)


ABaldwinHunter:
In Classic, we used `effective_mass`, which was the node_size for similar code, and node_size * num_occurrences for identical, or essentially:

`flay_score / num_occurrences`

pbrisbin:
> [F]lay score is:
>
> Similar code
> `node_size * number_of_occurrences_of_duplicated_issue`
> 
> Identical code
> `node_size * num_occurrences * num_occurrences`
>
> In Classic, we used ... `node_size` for similar code, and `node_size * num_occurrences` for identical, or essentially:
>
> flay_score / num_occurrences

I'm having a little trouble matching your commit/PR description to what's going on.

I've I'm reading the diff correctly, what's really going on is that we're going to use `Sexp#mass` instead of `Flay::Item#mass` (?). Additionally, `Flay::Item#mass` was the "Flay" equations you show and `Sexp#mass` is the New `flay_score / num_occurrences` equation -- is all that right?

ABaldwinHunter:
@pbrisbin Almost all right.

flay score is indeed `Flay::Item#mass`
```
 Sexp#mass * num_occurrences (similar code)
 Sexp#mass * num_occurrences **2 (identical code)
```

Duplication engine had previously used flay score for `mass` when reporting mass in UI, and doing remediation points calculations.

[Classic](https://github.com/codeclimate/analyzer/blob/master/lib/quality/worker/jobs/smell_duplication.rb#L29): used flay_score / num_occurrences (effectively), or
```
Sexp#mass (similar code)
Sexp#mass * num_occurrences (identical code)
```

This update: straight up `Sexp#mass` always

ABaldwinHunter:
@pbrisbin ready for another look!

- passed boolean attribute `identical` to `Violation` instead of derivative string `check_name` 
- clarified commit message. for easy ref:
```
Use Sexp#mass for issue mass, not Flay::Item#mass

Better matches Classic.

Additionally, stop penalizing extra for duplicated code that is identical,
as its remediation effort should be the same or less than that of similar code.
```

pbrisbin:
LGTM

PR #86, opened 2016-01-27
Tune Python Remediation Points

ABaldwinHunter:
- Reduce AST threshold from 40 to 32 (classic is 28)
- update point formula to match classic computation

Change from
     remediations_points = 1000 * mass
to
     remediation_points = 1_000_000 + (mass - threshold)  * 50_000

@codeclimate/review 

Third starfish leg of https://github.com/codeclimate/codeclimate-duplication/pull/78/files

### DIFF
--- a/lib/cc/engine/analyzers/python/main.rb
+++ b/lib/cc/engine/analyzers/python/main.rb
@@ -12,10 +12,19 @@ module CC
         class Main < CC::Engine::Analyzers::Base
           LANGUAGE = "python"
           DEFAULT_PATHS = ["**/*.py"]
-          DEFAULT_MASS_THRESHOLD = 40
-          BASE_POINTS = 1000
+          DEFAULT_MASS_THRESHOLD = 32
+          BASE_POINTS = 1_500_000
+          POINTS_PER_OVERAGE = 50_000
+
+          def calculate_points(mass)
+            BASE_POINTS + (overage(mass) * POINTS_PER_OVERAGE)
+          end
 
           private
+
+          def overage(mass)
+            mass - mass_threshold
+          end
 
           def process_file(path)
             Node.new(::CC::Engine::Analyzers::Python::Parser.new(File.binread(path), path).parse.syntax_tree, path).format
--- a/spec/cc/engine/analyzers/python/main_spec.rb
+++ b/spec/cc/engine/analyzers/python/main_spec.rb
@@ -28,7 +28,7 @@ print("Hello", "python")
         "path" => "foo.py",
         "lines" => { "begin" => 1, "end" => 1 },
       })
-      expect(json["remediation_points"]).to eq(6_000)
+      expect(json["remediation_points"]).to eq(1_600_000)
       expect(json["other_locations"]).to eq([
         {"path" => "foo.py", "lines" => { "begin" => 2, "end" => 2} },
         {"path" => "foo.py", "lines" => { "begin" => 3, "end" => 3} }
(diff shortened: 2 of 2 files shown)


ABaldwinHunter:
cc @pbrisbin

lib/cc/engine/analyzers/python/main.rb
```diff
@@ -12,11 +12,20 @@ module Python
...
+          BASE_POINTS = 1_500_000
+          POINTS_PER_OVERAGE = 50_000
+
+          def calculate_points(mass)
+            BASE_POINTS + (overage(mass) * POINTS_PER_OVERAGE)
+          end
 
           private
 
+          def overage(mass)
```
pbrisbin: What do you think about assigning a local `overage` variable directly in `#mass`, rather than the method call. Names are good, but extra indirection can be bad, and the fact that you have to thread the `mass` variable down to this private method is a point against IMO.

lib/cc/engine/analyzers/python/main.rb
```diff
@@ -12,11 +12,20 @@ module Python
...
+            mass - mass_threshold
```
pbrisbin: Where's `mass_threshold`? Did you mean `DEFAULT_MASS_THRESHOLD` here?

ABaldwinHunter: No, you can also set a custom threshold.

The method is inherited. 
https://github.com/codeclimate/codeclimate-duplication/blob/master/lib/cc/engine/analyzers/analyzer_base.rb#L35

lib/cc/engine/analyzers/python/main.rb
ABaldwinHunter: You mean in `#calculate_points`?

That's fine by me...

I think it should be a separate refactor though.

So far, the ruby one uses the same format.

ABaldwinHunter: https://github.com/codeclimate/codeclimate-duplication/blob/master/lib/cc/engine/analyzers/ruby/main.rb#L31

lib/cc/engine/analyzers/python/main.rb
pbrisbin: I see. OK.

lib/cc/engine/analyzers/python/main.rb
pbrisbin: Sounds good.

pbrisbin:
LGTM

lib/cc/engine/analyzers/python/main.rb
pbrisbin: **Something to fix later, not to hold up this PR...**

Wait, why do these methods exist more than once? Can't they be shared on `Base` then?

ABaldwinHunter: Yes +1

I guess the overage method could already by shared  innocuously.

The `calculate_points` method hasn't been moved yet because 2 of the 4 languages use the old `calculate_points` formula (base * mass)

pbrisbin: > The calculate_points method hasn't been moved yet because 2 of the 4 languages use the old calculate_points formula (base * mass)

Is that intentional? Is the card you're working on fix "Duplication" or fix "Python"?

ABaldwinHunter: ATM I'd say `Fix python-duplication` which is the child of `Fix python`.

We've been tackling grade tuning on a per-language basis, descending into the language's engines.

Not sure if that's the best approach.

pbrisbin: OK, sounds good.

PR #88, opened 2016-01-27
Minor refactor

ABaldwinHunter:
Clean specs, remove unused method.

@codeclimate/review

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,4 +1,5 @@
 require 'bundler/setup'
+require 'flay'
 require 'tmpdir'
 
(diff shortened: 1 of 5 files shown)


spec/spec_helper.rb
```diff
@@ -1,6 +1,8 @@
 require 'bundler/setup'
+require 'flay'
 require 'tmpdir'
 
+
```
pbrisbin: Intentional?

pbrisbin:
LGTM, modulo test failures.

spec/spec_helper.rb
ABaldwinHunter: nope, good catch. thanks

PR #92, opened 2016-01-29
Fix fingerprint bug - don't use num occurrences

ABaldwinHunter:
Using the number of occurrences of duplicated code as a unique
identifier of the issue makes the fingerprint non-deterministic
with respect to a given issue at a site of the duplicated code.

Opting to use checkname instead.

Fixes false positives in PR comparisons: https://codeclimate.com/repos/569fa9e70c40f068790031e2/pull/3#duplication

cc @codeclimate/review

### DIFF
--- a/lib/cc/engine/analyzers/violation.rb
+++ b/lib/cc/engine/analyzers/violation.rb
@@ -86,7 +86,7 @@ module CC
           digest << "-"
           digest << current_sexp.mass.to_s
           digest << "-"
-          digest << occurrences.to_s
+          digest << check_name
           digest.to_s
         end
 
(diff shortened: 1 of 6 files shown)


GordonDiggs:
LGTM

lib/cc/engine/analyzers/violation.rb
```diff
@@ -86,7 +86,7 @@ def fingerprint
           digest << "-"
           digest << current_sexp.mass.to_s
           digest << "-"
-          digest << occurrences.to_s
+          digest << check_name
```
wfleming: Classic also uses the type of the sexp's root node, I think. That's an extra piece of info to make the fingerprint more unique that may be worth adding.

PR #93, opened 2016-02-01
Handle Open3 returning a nil status

pbrisbin:
Customers are experiencing this in production. It results in an engine error
output of:

    NoMethodError: undefined method `success?' for nil:NilClass
        block in run at /usr/src/app/lib/cc/engine/analyzers/command_line_runner.rb:18

The included spec reproduces this error.

It's unclear what circumstances lead to this, but removing the NoMethodError
will at least let the engine's stderr come through, which we can then debug and
address thereafter.

/cc @codeclimate/review

I also added basic spec coverage before beginning, as its own commit.

### DIFF
--- a/lib/cc/engine/analyzers/command_line_runner.rb
+++ b/lib/cc/engine/analyzers/command_line_runner.rb
@@ -15,6 +15,9 @@ module CC
         def run(input)
           Timeout.timeout(timeout) do
             out, err, status = Open3.capture3(command, stdin_data: input)
+
+            status ||= handle_open3_race_condition(out)
+
             if status.success?
               yield out
             else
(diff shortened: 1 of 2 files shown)


lib/cc/engine/analyzers/command_line_runner.rb
```diff
@@ -15,10 +15,11 @@ def initialize(command, timeout = DEFAULT_TIMEOUT)
         def run(input)
           Timeout.timeout(timeout) do
             out, err, status = Open3.capture3(command, stdin_data: input)
-            if status.success?
+            if status && status.success?
               yield out
             else
-              raise ::CC::Engine::Analyzers::ParserError, "`#{command}` exited with code #{status.exitstatus}:\n#{err}"
+              exitstatus = status && status.exitstatus
+              raise ::CC::Engine::Analyzers::ParserError, "`#{command}` exited with code #{exitstatus}:\n#{err}"
```
GordonDiggs: Should this be `exitstatus.inspect` so that we get the `nil` explicitly printed?

wfleming:
FWIW I looked into this a bit a while back & from my notes, I believed this wasn't actually a failure case, but a race condition: the `Open3` code internally uses `Process.detach` to get the wait thread, which returns `nil` if the pid given to it is not a valid pid. So my guess was that the parser ran & exited so quickly that by the time `detach` was called it didn't exist anymore. But it may very well have ran successfully.

My thinking was we should move away from `open3` and use the same posix spawn gem patterns we do elsewhere (if it's not reliant on C bindings).

pbrisbin:
I'll happily move to `POSIX::Spawn`, will ping when ready for re-review.

pbrisbin:
Changed, squashed, and commit message updated. Ready for re-review.

wfleming:
LGTM.

pbrisbin:
Had to go back to the original implementation because posix-spawn's not supported on JRuby.

Please see the last commit for a detailed message: https://github.com/codeclimate/codeclimate-duplication/commit/ca5d18723fd53da9d65a0cee1cf55132cba7431f

wfleming:
I'm cool with it.

PR #94, opened 2016-02-01
Update penalty algorithm for PHP and JavaScript

ABaldwinHunter:
Bug fix to make penalty calculation harsher. Matches classic.

Refactor methods and defaults to make use of base class.

Additional tuning of threshold (and `per_overage_points` values) for PHP and JavaScript to come. 

On classic, base = 1_500_000, per = 50_000, for both.
For now, I made per_point value proportional to threshold.

An important consequence of this change is that grades will drop. 

@codeclimate/review

### DIFF
--- a/lib/cc/engine/analyzers/javascript/main.rb
+++ b/lib/cc/engine/analyzers/javascript/main.rb
@@ -16,7 +16,7 @@ module CC
           ]
           LANGUAGE = "javascript"
           DEFAULT_MASS_THRESHOLD = 40
-          BASE_POINTS = 3000
+          POINTS_PER_OVERAGE = 30_000
 
           private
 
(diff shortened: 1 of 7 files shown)


lib/cc/engine/analyzers/javascript/main.rb
```diff
@@ -16,7 +16,7 @@ class Main < CC::Engine::Analyzers::Base
           ]
           LANGUAGE = "javascript"
           DEFAULT_MASS_THRESHOLD = 40
-          BASE_POINTS = 3000
+          POINTS_PER_OVERAGE = 30_000
```
pbrisbin: Why is this necessary to do, in addition to just fixing the equation(s)?

ABaldwinHunter: The points per overage vary by language (roughly a function of mass_threshold / granularity of node parsing).

Since the BASE_POINTS and the formula are the same, they can be refactored to live in the Base Class.

pbrisbin:
LGTM

PR #95, opened 2016-02-01
Add pending spec for over-escaped multibyte source

pbrisbin:
Some customers have code with escaped multibyte characters in regular
expressions. This causes the engine to error because File.read doubly-escapes
it, which fails to parse.

The escaping these customers have is not needed. They can remove their
backslash and all should work fine. That's what we're prescribing for now via
customer services, since all other options are too time-consuming.

Eventually, we would like to support this code as-is (as we do any realistic
source files customers may have) -- when/if we do, this spec will pass.

/cc @codeclimate/review


pbrisbin:
Ping @wfleming

ABaldwinHunter:
LGTM

PR #102, opened 2016-02-19
bump ruby_parser & sexp_processor

wfleming:
New releases today with Ruby 2.3 support. I guess the ones we updated to yesterday didn't have all of that support, or maybe I misread commit logs.

FYI @codeclimate/review


jpignata:
:100:

PR #109, opened 2016-03-22
Add debug logging option

dblandin:
This PR adds a configuration option for debug logging in order to shake out some difficult to diagnose analysis issues.

@codeclimate/review :mag_right:

### DIFF
--- a/lib/cc/engine/analyzers/reporter.rb
+++ b/lib/cc/engine/analyzers/reporter.rb
@@ -16,8 +16,12 @@ module CC
         end
 
         def run
+          debug("Processing #{language_strategy.files.count} files concurrency=#{engine_config.concurrency}")
+
           process_files
           report
+
+          debug("Reported #{reports.size} violations...")
         end
 
         def process_files
@@ -26,12 +30,20 @@ module CC
             concurrency: engine_config.concurrency
           )
 
+          processed_files_count = 0
+
           pool.run do |file|
+            debug("Processing file: #{file}")
+
             sexp = language_strategy.run(file)
             process_sexp(sexp)
+
+            processed_files_count += 1
           end
 
           pool.join
+
+          debug("Processed #{processed_files_count} files")
         end
 
         def report
@@ -39,6 +51,8 @@ module CC
             violations = new_violations(issue)
 
             violations.each do |violation|
+              debug("Violation name=#{violation.report_name} mass=#{violation.mass}")
+
               unless reports.include?(violation.report_name)
                 reports.add(violation.report_name)
                 io.puts "#{violation.format.to_json}\0"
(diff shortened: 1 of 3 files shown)


lib/cc/engine/duplication.rb
```diff
@@ -30,6 +30,10 @@ def initialize(directory:, engine_config:, io:)
       def run
         Dir.chdir(directory) do
           languages_to_analyze.each do |language|
+            if engine_config.debug?
+              $stderr.puts("Analyzing language #{language}")
```
pbrisbin: This seems to be the only log line that can't use `Reporter#debug` and has to duplicate the logic. Thoughts on just ditching it? Seems low value to me.

pbrisbin:
LGTM. Note that to do CLI testing, you'll need to make a change there to display engine STDERR even in the successful case (it doesn't do this currently).

lib/cc/engine/duplication.rb
dblandin: Ditching it sounds good to me.

dblandin:
> Note that to do CLI testing, you'll need to make a change there to display engine STDERR even in the successful case (it doesn't do this currently).

Right. I've been running the engine directly with `docker run` to test. For our purposes debugging some specific issues, I think it's fine if it doesn't show up during CLI runs. We just need to see the engine stderr on codeclimate.com.

Do you think this is useful enough generally to make that CLI change?

lib/cc/engine/analyzers/reporter.rb
```diff
@@ -26,12 +30,20 @@ def process_files
...
 
+          processed_files_count = 0
+
           pool.run do |file|
+            debug("Processing file: #{file}")
+
             sexp = language_strategy.run(file)
             process_sexp(sexp)
+
+            processed_files_count += 1
```
GordonDiggs: Do you need anything here to handle the concurrency and incrementing this number?

dblandin: I don't think so. The join below should block until all threads return.

GordonDiggs: Right but this block is executed inside the thread right? So more than one thread could try to increment this at once

pbrisbin:
> Do you think this is useful enough generally to make that CLI change?

Nope, just letting you know about a gotcha that I've hit before. Whoever thinks it's useful should make that change.

lib/cc/engine/analyzers/reporter.rb
dblandin: Gotcha, I suppose that's true! I'll wrap the increment with a mutex.

GordonDiggs: You could also use this class from concurrent-ruby I think http://ruby-concurrency.github.io/concurrent-ruby/Concurrent/AtomicFixnum.html

dblandin: Nice! That's handy!

PR #111, opened 2016-03-24
Ensure FileList never emits directories

pbrisbin:
Apparently it's common for JavaScript projects to have a path like:

    ./foo.js/foo.src.js

Since the directory matches `**/*.js`, directory expansions emits it as a file
and the analyzer errors (`EISDIR`) trying to read it.

The solution is to ensure we only ever emit files from the glob-expansion. The
additional guard in `#matches?` is not strictly necessary but was added to
increase robustness in case upstream logic changes and the entries in
include-paths become unreliable as to their file-vs-directory-ness.

/cc @codeclimate/review

### DIFF
--- a/spec/cc/engine/analyzers/file_list_spec.rb
+++ b/spec/cc/engine/analyzers/file_list_spec.rb
@@ -42,5 +42,20 @@ RSpec.describe CC::Engine::Analyzers::FileList do
 
       expect(file_list.files).to eq(["./foo.js"])
     end
+
+    it "does not emit directories even if they match the patterns" do
+      file_list = ::CC::Engine::Analyzers::FileList.new(
+        engine_config: CC::Engine::Analyzers::EngineConfig.new(
+          "include_paths" => ["./"],
+        ),
+        patterns: ["**/*.js"],
+      )
+
+      Dir.mkdir("vendor.js")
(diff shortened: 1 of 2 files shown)


spec/cc/engine/analyzers/file_list_spec.rb
```diff
@@ -42,5 +42,19 @@
...
+        engine_config: CC::Engine::Analyzers::EngineConfig.new(
+          "include_paths" => ["./"],
+        ),
+        patterns: ["**/*.js"],
+      )
+
+      Dir.mkdir("vendor.js")
+      File.write(File.join(@tmp_dir, "vendor.js", "vendor.src.js"), "")
+
+      expect(file_list.files).not_to include("./vendor.js")
```
wfleming: Might be worth asserting here that the list *does* include `./vendor.js/vendor.src.js`?

ABaldwinHunter: +1

ABaldwinHunter:
I like @wfleming's test idea.

Other than that, LGTM. Nice fix

spec/cc/engine/analyzers/file_list_spec.rb
pbrisbin: Done.

PR #114, opened 2016-03-31
Mention "What is duplication?" in README

wfleming:
Some people come to the repo instead of visiting docs. Linking to that
page here so people can find it.

cc @codeclimate/review


pbrisbin:
Sounds good to me.

Issue #115, opened 2016-05-04
Excluding files and/or directories


Mange:
Test code is sometimes marked as duplicate and/or similar, which is usually not a problem for me since I accept some duplication in tests just to get clearer and more focused expectations.

It would be helpful to be able to ignore certain files or directories from this engine so I don't get those kinds of checks done on the tests.

Is that possible?

devinus:
Yeah, I need to ignore all my tests as they're almost always duplicated code.

jekuta:
You can exclude files and folders for specific engines. I raised this same issue a little while ago and got an answer here. https://github.com/codeclimate/codeclimate-duplication/issues/113

Mange:
Thanks, Laura.

Could that information be added to the documentation, perhaps?

Den ons 4 maj 2016 08:34Laura Paakkinen <notifications@github.com> skrev:

pbrisbin:
Hi @Mange that information is present in our main documentation site:

https://docs.codeclimate.com/docs/excluding-files-and-folders#section-exclude-paths-for-specific-engines

If you have any suggestions for improving those docs (or making them easier to find), please let us know.

Mange:
Thanks. I opened a PR with my suggested changes. :-)

PR #116, opened 2016-05-06
Mention possibility to exclude files

Mange:
The option is part of the "global settings", but what those global settings are and that they exist isn't obvious from the README. The documentation shown when clicking on duplication errors on `codeclimate.com` also does not mention this at all.

This is probably a good entry point for people looking for it.

Relates to #115.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,10 @@ the Hash type to support extra configuration.
 
 ### Excluding files and directories
 
-As with any other Codeclimate engine, you can exclude certain files or directories from being analyzed. For more information, [see *Exclude paths for specific engines* on our documentation][exclude-files-engine].
+As with any other Code Climate engine, you can exclude certain files or
+directories from being analyzed. For more information, see
+[*Exclude paths for specific engines*][exclude-files-engine] in our
+documentation.
 
 ```yaml
 engines:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -68,7 +68,19 @@ engines:
 Note that you have the update the YAML structure under the `languages` key to
 the Hash type to support extra configuration.
 
+### Excluding files and directories
+
+As with any other Codeclimate engine, you can exclude certain files or directories from being analyzed. For more information, [see *Exclude paths for specific engines* on our documentation][exclude-files-engine].
```
pbrisbin: Couple nits:

"Code Climate", not "Codeclimate". Can you hard-wrap this at 80 columns, like the rest of the document? I think just the bit you've emphasized (Exclude paths for specific engines) should be the link (rather than emphasizing it within a bigger link).

pbrisbin:
Just some small copy suggestions, then I'll get this merged. Thanks!

README.md
Mange: That makes sense. I tried to get the link to be as descriptive as possible for a17n, while emphasizing the subtitle of the target document.

Mange:
Is that better? :-)

pbrisbin:
Perfect, thanks!

PR #117, opened 2016-05-19
Abort when an exception is raised

ABaldwinHunter:
This change fixes behavior where a JVM OOM exception sometimes did not result
in the engine raising an error and crashing.

### DIFF
--- a/lib/cc/engine/analyzers/file_thread_pool.rb
+++ b/lib/cc/engine/analyzers/file_thread_pool.rb
@@ -7,6 +7,8 @@ module CC
         DEFAULT_CONCURRENCY = 2
         MAX_CONCURRENCY = 2
 
+        Thread.abort_on_exception = true
+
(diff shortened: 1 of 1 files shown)


ABaldwinHunter:
Examining failing test.

pbrisbin:
This probably good to do anyway, but I'm not sure it'll solve the problem. I don't know the relationship between Ruby's `Thread` class and these JVM threads that are dying. Did you find any documentation or have some kind of reproduction case that makes you confident in this as a fix?

ABaldwinHunter:
@pbrisbin Currently digging in a bit and exploring the [thread docs](http://ruby-doc.org/core-2.2.0/Thread.html).

A repro case also sounds helpful - I can go ahead and build that.

dblandin:
@ABaldwinHunter Is this PR still active?

dblandin:
Going to close for now.

ABaldwinHunter:
👍 thanks @dblandin

PR #124, opened 2016-07-18
Use the spreadRestParams plugin for Babylon

wfleming:
From CX escalation: parsing spread params (e.g. `...foo`) in Babylon
requires enabling a plugin.

cc @codeclimate/review


GordonDiggs:
LGTM

PR #125, opened 2016-07-18
Reorder npm install/package.json copy

wfleming:
This treats this more like other dependencies, and stops us from doing
`npm install` on every single build.

cc @codeclimate/review


GordonDiggs:
LGTM

PR #130, opened 2016-09-22
Tweak JVM config to increase heap size

ABaldwinHunter:
- Tune heap parameters
  We have seen `java.lang.OutOfMemoryError: Java heap space` errors
  The default heap size is 512mb. Increase to 1024.

  Overview of Heap tuning options:
  https://docs.oracle.com/cd/E19159-01/819-3681/6n5srlhqp/index.html

  The min and max heap ratios determine the min and max amount of free space
  permitted between garbage collections, as a percentage. The defaults are
  40 and 70 resp, according to
  http://www.avricot.com/blog/?post/2010/05/03/Get-started-with-java-JVM-memory-(heap%2C-stack%2C-xss-xms-xmx-xmn...)

- Use Parallel GC instead of ParNew which is deprecated when used with the
  SerialOld Parser.

  See: http://openjdk.java.net/jeps/173

### DIFF
--- a/lib/cc/engine/analyzers/analyzer_base.rb
+++ b/lib/cc/engine/analyzers/analyzer_base.rb
@@ -10,6 +10,7 @@ module CC
           ::Errno::ENOENT,
           ::Racc::ParseError,
           ::RubyParser::SyntaxError,
+          ::RuntimeError,
         ].freeze
 
         BASE_POINTS = 1_500_000
--- a/lib/cc/engine/analyzers/analyzer_base.rb
+++ b/lib/cc/engine/analyzers/analyzer_base.rb
@@ -10,7 +10,6 @@ module CC
           ::Errno::ENOENT,
           ::Racc::ParseError,
           ::RubyParser::SyntaxError,
-          ::RuntimeError,
         ].freeze
 
         BASE_POINTS = 1_500_000
--- a/lib/cc/engine/analyzers/file_thread_pool.rb
+++ b/lib/cc/engine/analyzers/file_thread_pool.rb
@@ -7,13 +7,14 @@ module CC
         DEFAULT_CONCURRENCY = 2
         MAX_CONCURRENCY = 2
 
+        Thread.abort_on_exception = true
+
         def initialize(files, concurrency: DEFAULT_CONCURRENCY)
           @files = files
           @concurrency = concurrency
         end
 
         def run(&block)
-          Thread.abort_on_exception = true
           queue = build_queue
           lock = Mutex.new
 
@@ -24,8 +25,6 @@ module CC
               end
             end
           end
-        ensure
-          Thread.abort_on_exception = false
         end
 
         def join
--- a/lib/cc/engine/analyzers/file_thread_pool.rb
+++ b/lib/cc/engine/analyzers/file_thread_pool.rb
@@ -7,6 +7,8 @@ module CC
         DEFAULT_CONCURRENCY = 2
         MAX_CONCURRENCY = 2
 
+        Thread.abort_on_exception = true
+
         def initialize(files, concurrency: DEFAULT_CONCURRENCY)
           @files = files
           @concurrency = concurrency
--- a/lib/cc/engine/analyzers/file_thread_pool.rb
+++ b/lib/cc/engine/analyzers/file_thread_pool.rb
@@ -7,14 +7,13 @@ module CC
         DEFAULT_CONCURRENCY = 2
         MAX_CONCURRENCY = 2
 
-        Thread.abort_on_exception = true
-
         def initialize(files, concurrency: DEFAULT_CONCURRENCY)
           @files = files
           @concurrency = concurrency
         end
 
         def run(&block)
+          Thread.abort_on_exception = true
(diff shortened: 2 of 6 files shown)


ABaldwinHunter:
@codeclimate/review 

@pbrisbin I'm curious about your thoughts after your work in https://github.com/codeclimate/codeclimate-duplication/commit/63dde4bb831a13b62161289385b2c4ddbcc4ddf9.

(That seems like it referred to docker container OOM errors, correct?)

I've been doing some local testing, and have also considered releasing the engine as beta first. What other testing might we want to see to have confidence this change performs well across a variety of repos?

lib/cc/engine/analyzers/analyzer_base.rb
```diff
@@ -10,7 +10,6 @@ class Base
           ::Errno::ENOENT,
           ::Racc::ParseError,
           ::RubyParser::SyntaxError,
-          ::RuntimeError,
```
wfleming: This seems pretty unrelated to this change?

This was added as a [bugfix](https://github.com/codeclimate/codeclimate-duplication/commit/6715ad152d95c3533e2722b3fd6e716354fb3616) because some ruby parser errors are thrown as `RuntimeException` instances, and we want to catch those.

It's very unusual for thrown exceptions to be `RuntimeError` instances: usually such errors would inherit from `RuntimeError`, which is why the `rescue` clause in `#run` will re-raise errors that are not this exact class.

ABaldwinHunter: @wfleming ah okay, I may have misclassified this error. What else might have prevented this error from halting analysis?:
```
Exception in thread "Ruby-0-Thread-3: /usr/src/app/lib/cc/engine/analyzers/file_thread_pool.rb:20" java.lang.OutOfMemoryError: Java heap space
	at java.nio.HeapCharBuffer.<init>(HeapCharBuffer.java:57)
	at java.nio.CharBuffer.allocate(CharBuffer.java:335)
	at org.jruby.RubyEncoding$UTF8Coder.<init>(RubyEncoding.java:304)
	at org.jruby.RubyEncoding.getUTF8Coder(RubyEncoding.java:365)
	at org.jruby.RubyEncoding.decodeUTF8(RubyEncoding.java:280)
	at org.jruby.runtime.Helpers.decodeByteList(Helpers.java:2642)
	at org.jruby.RubyString.decodeString(RubyString.java:759)
	at org.jruby.RubyString.toString(RubyString.java:750)
	at org.jruby.RubyString.asJavaString(RubyString.java:1276)
	at usr.src.app.lib.cc.engine.analyzers.php.ast.RUBY$block$json_to_ast$7(/usr/src/app/lib/cc/engine/analyzers/php/ast.rb:40)
```

wfleming: I don't know: I think we've seen that before, and the suspicion would be somewhere around `FileThreadPool`. I don't think the catch here is involved, since I don't see this file in that stacktrace. Is there a log message near that exception that looks like [the one the catch here logs](https://github.com/codeclimate/codeclimate-duplication/blob/42a53aece30e5f51c251fe2c4e59b501ccb9c225/lib/cc/engine/analyzers/analyzer_base.rb#L24)?

ABaldwinHunter: Ah you know what, I think you're right. This rescue isn't related.

I think you're right about the ThreadPool issue - I think we may not be raising on errors in child threads. 

Maybe I should bring back this attempt to change that behavior: https://github.com/codeclimate/codeclimate-duplication/pull/117

lib/cc/engine/analyzers/file_thread_pool.rb
```diff
@@ -7,6 +7,8 @@ class FileThreadPool
         DEFAULT_CONCURRENCY = 2
         MAX_CONCURRENCY = 2
 
+        Thread.abort_on_exception = true
```
wfleming: Do you have any kind of reproduction case that can give us some confidence this will help? [Thread#join](http://ruby-doc.org/core-2.2.0/Thread.html#method-i-join) should raise any exceptions when this isn't set. Do we have reason to think `#join` is not working as intended in JRuby/this engine?

ABaldwinHunter: Working on reproduction case

ABaldwinHunter: @wfleming: @maxjacobson and I confirmed that this change causes the OOM java error to get raised in the ruby thread. We used a toy example for testing and `Thread#join` didn't cause the exception to make it through.

errored build: https://codeclimate.com/repos/57e4d7aad0d6c36836004456/builds/9
beta channel branch: https://github.com/codeclimate/codeclimate-duplication/tree/channel/beta

Updating this branch

ABaldwinHunter:
After some additional experimentation, I don't have confidence in the JAVA_OPTS proposed above. I'm going to close this PR and open the error raising and (potentially) JAVA_OPTS tweak PRs separately. 

Closing in favor of https://github.com/codeclimate/codeclimate-duplication/pull/132

PR #132, opened 2016-09-29
Raise OOM error

ABaldwinHunter:
@codeclimate/review 

Abort when exception is raised during analysis
Fixes an issue where Java OOM exceptions were swallowed and could result in
issues getting dropped.

We suppress the thread abort behavior in the CommandLine Runner class to
ensure we get a meaningful error whene there's a timeout, instead of an
IOError.

See individual commits for details on other changes.

Thanks to @maxjacobson for help confirming this behavior. 

Closes https://github.com/codeclimate/codeclimate-duplication/pull/130/files

(The JAVA_OPTS tuning isn't quite right in pull 130 and should be addressed separately).

### DIFF
--- a/lib/cc/engine/analyzers/file_thread_pool.rb
+++ b/lib/cc/engine/analyzers/file_thread_pool.rb
@@ -16,8 +16,8 @@ module CC
           queue = build_queue
           lock = Mutex.new
 
-          @workers = thread_count.times.map do
-            Thread.new do
+          @workers = Array.new(thread_count) do
+            with_thread_abortion do
               while (item = next_item(queue, lock))
                 yield item
               end
@@ -53,6 +53,13 @@ module CC
           else
             DEFAULT_CONCURRENCY
           end
+        end
+
+        def with_thread_abortion
+          t = Thread.new do
+            yield
+          end
+          (t.abort_on_exception = true) && t
         end
       end
     end
--- a/circle.yml
+++ b/circle.yml
@@ -20,7 +20,7 @@ test:
 
 deployment:
   registry:
-    branch: master
+    branch: /master|channel\/[\w-]+/
     commands:
       - >
         docker run
--- a/spec/cc/engine/analyzers/file_thread_pool_spec.rb
+++ b/spec/cc/engine/analyzers/file_thread_pool_spec.rb
@@ -3,8 +3,9 @@ require "cc/engine/analyzers/file_thread_pool"
 
 RSpec.describe CC::Engine::Analyzers::FileThreadPool do
   describe "#run" do
+    let(:thread) { Thread.new {} }
     it "uses default count of threads when concurrency is not specified" do
-      allow(Thread).to receive(:new)
+      allow(Thread).to receive(:new).and_return(thread)
 
       pool = CC::Engine::Analyzers::FileThreadPool.new([])
       pool.run  {}
@@ -15,7 +16,7 @@ RSpec.describe CC::Engine::Analyzers::FileThreadPool do
     end
 
     it "uses default concurrency when concurrency is over max" do
-      allow(Thread).to receive(:new)
+      allow(Thread).to receive(:new).and_return(thread)
 
       run_pool_with_concurrency(
         CC::Engine::Analyzers::FileThreadPool::DEFAULT_CONCURRENCY + 2
@@ -27,7 +28,7 @@ RSpec.describe CC::Engine::Analyzers::FileThreadPool do
     end
 
     it "uses default concucurrency when concucurrency is under 1" do
-      allow(Thread).to receive(:new)
+      allow(Thread).to receive(:new).and_return(thread)
 
       run_pool_with_concurrency(-2)
 
@@ -37,7 +38,7 @@ RSpec.describe CC::Engine::Analyzers::FileThreadPool do
     end
 
     it "uses supplied concurrency when valid" do
-      allow(Thread).to receive(:new)
+      allow(Thread).to receive(:new).and_return(thread)
 
       run_pool_with_concurrency(1)
 
@@ -56,6 +57,14 @@ RSpec.describe CC::Engine::Analyzers::FileThreadPool do
       expect(results).to include("cba")
       expect(results).to include("321")
       expect(results).to include("zyx")
+    end
+
+    it "aborts on a thread exception" do
+      allow(Thread).to receive(:new).and_return(thread)
+
+      run_pool_with_concurrency(1)
+
+      expect(thread.abort_on_exception).to eq(true)
     end
   end
 


brynary:
Whoa, scary PR :ghost:

lib/cc/engine/analyzers/file_thread_pool.rb
```diff
@@ -16,12 +16,13 @@ def run(&block)
...
 
-          @workers = thread_count.times.map do
-            Thread.new do
+          @workers = Array.new(thread_count) do
+            t = Thread.new do
               while (item = next_item(queue, lock))
                 yield item
               end
             end
+            t.abort_on_exception = true
```
maxjacobson: ah this is a nice improvement. I think you need to add another line after this one that just returns `t` though

ABaldwinHunter: thanks max. good call

Dockerfile
```diff
@@ -27,6 +27,7 @@ RUN chown -R app .
 
 USER app
 
-ENV JAVA_OPTS="-XX:+UseParNewGC -XX:MinHeapFreeRatio=5 -XX:MaxHeapFreeRatio=10"
```
maxjacobson: did you mean to remove the existing opts or just undo your proposed change to them?

ABaldwinHunter:
For reference this issue seems relevant: https://github.com/jruby/jruby/issues/2354

Makefile
```diff
@@ -6,7 +6,7 @@ image:
 	docker build --rm -t $(IMAGE_NAME) .
 
 test: image
-	docker run --tty --interactive --rm $(IMAGE_NAME) bundle exec rake
+	docker run --tty --interactive --rm $(IMAGE_NAME) sh -c "cd /usr/src/app && bundle exec rake"
```
pbrisbin: I'm curious why this was needed, were the tests not running before?

You can use

```
docker run --tty --interactive --rm --workdir /usr/src/app $(IMAGE_NAME) bundle exec rake
```

Instead of jumping to `sh` so you can `cd` as part of the command.

lib/cc/engine/analyzers/file_thread_pool.rb
pbrisbin: I guess this is more localized, but what's the downside of just putting `Thread.abort_on_exception = true` somewhere at startup?

Makefile
pbrisbin: Ah, I see why this is needed now. My suggestion to use `--workdir` still stands.

pbrisbin:
> We suppress the thread abort behavior in the CommandLine Runner class to
ensure we get a meaningful error whene there's a timeout, instead of an
IOError.

I was surprised and curious about this, but I also don't see any code in here related to it -- is this message out of date? If not, can you explain?

ABaldwinHunter:
@pbrisbin sorry that's left over from an earlier implementation. updating

Dockerfile
wfleming: It looks like the old options are now on line 33.

lib/cc/engine/analyzers/file_thread_pool.rb
brynary: @ABaldwinHunter RE: @pbrisbin's comment... is it the case that we want different `abort_on_exception` on different threads within this process?

Dockerfile
ABaldwinHunter: yeah, thanks for the catch! I think actually did make that mistake but github doesn't seem to squash the comments anymore :(

Makefile
ABaldwinHunter: 👍  updated

lib/cc/engine/analyzers/file_thread_pool.rb
ABaldwinHunter: @brynary BTW I looked at the `thread` object returned from `Thread#join` but so far didn't see a quick and useful way to interrogate whether there'd been an exception. I also think there could be some slight performance advantages to aborting early.

WDYT about the current implementation instead?

Dockerfile
ABaldwinHunter: Hm actually I guess it makes sense since the deleted line remains deleted

maxjacobson: Yea. I do feel like there should be a button in this new review system where we can mark a thread as resolved and collapse it

Makefile
ABaldwinHunter: @pbrisbin Actually, I removed the docker file changes. They fall into category of driveby and also independent of any other change here result in an exception getting raised that looks like:

```
LoadError: load error: bundler/setup -- java.lang.RuntimeException: Unhandled IOException: java.io.IOException: unhandled errno: Read-only file system
  require at org/jruby/RubyKernel.java:939
  require at /opt/jruby/lib/ruby/stdlib/rubygems/core_ext/kernel_require.rb:128
    <top> at /usr/src/app/lib/cc/engine/duplication.rb:1
  require at org/jruby/RubyKernel.java:939
   (root) at /opt/jruby/lib/ruby/stdlib/rubygems/core_ext/kernel_require.rb:1
    <top> at /opt/jruby/lib/ruby/stdlib/rubygems/core_ext/kernel_require.rb:54
```

ABaldwinHunter:
@codeclimate/review Ready for another look!

maxjacobson:
LGTM

lib/cc/engine/analyzers/file_thread_pool.rb
brynary: @ABaldwinHunter did you try `Thread#value` (I think it will be an exception)?

wfleming: @brynary Since `Thread#value` is meant to be implemented on top of `Thread#join`, I think it would suffer from the same bug where the exception wasn't raised. [This JRuby issue](https://github.com/jruby/jruby/issues/2354) is what's motivating this change.

ABaldwinHunter: @brynary thanks! Experimented with `Thread#value` in a toy example and did find some useful information. I don't think it will work for us here.

My observation: when `#value` is called on a thread that contains a ruby exception, the exception is raised immediately. This is similar to the way we had expected a thread exception to get raised on `#join`.

In our case here though, the nature of the Java Exception issue is that it doesn't end up raising a ruby error. It just prints a stacktrace and continues. See https://github.com/jruby/jruby/issues/2354

So, the thread that encounters the Java level exception doesn't end up with the desired `#value`.

In my toy example, all of the threads end up with `nil` value and the final `"made past all the threads" line still gets printed:

```
require 'thread'
require 'securerandom'

threads = []
400.times do |n|
  threads << Thread.new do
    if n == 350
      leak = []
      loop do
        leak << SecureRandom.hex(3000)
      end
    else
      puts n
    end
  end
end

p threads.map(&:value).uniq

puts "made it past all the threads"
```

=>

```
...
398
399
Exception in thread "Ruby-0-Thread-353: app.rb:1" java.lang.OutOfMemoryError: Java heap space
	at org.jruby.util.ConvertBytes.twosComplementToUnsignedBytes(ConvertBytes.java:213)
	at org.jruby.util.ConvertBytes.twosComplementToHexBytes(ConvertBytes.java:176)
	at org.jruby.ext.securerandom.SecureRandomLibrary.hex(SecureRandomLibrary.java:31)
	at org.jruby.ext.securerandom.SecureRandomLibrary$INVOKER$s$hex.call(SecureRandomLibrary$INVOKER$s$hex.gen)
	at org.jruby.runtime.callsite.CachingCallSite.call(CachingCallSite.java:161)
	at app.invokeOther1:hex(app.rb)
	at app.RUBY$block$\=app\,rb$2(app.rb:13)
	at java.lang.invoke.LambdaForm$DMH/1575540360.invokeStatic_L7_L(LambdaForm$DMH)
	at java.lang.invoke.LambdaForm$MH/1489069835.invokeExact_MT(LambdaForm$MH)
	at org.jruby.runtime.CompiledIRBlockBody.commonYieldPath(CompiledIRBlockBody.java:70)
	at org.jruby.runtime.IRBlockBody.yieldSpecific(IRBlockBody.java:67)
	at org.jruby.runtime.Block.yieldSpecific(Block.java:116)
	at org.jruby.RubyKernel.loop(RubyKernel.java:1291)
	at org.jruby.RubyKernel$INVOKER$s$0$0$loop.call(RubyKernel$INVOKER$s$0$0$loop.gen)
	at org.jruby.runtime.callsite.CachingCallSite.cacheAndCall(CachingCallSite.java:303)
	at org.jruby.runtime.callsite.CachingCallSite.callBlock(CachingCallSite.java:141)
	at org.jruby.runtime.callsite.CachingCallSite.call(CachingCallSite.java:145)
	at app.invokeOther5:loop(app.rb)
	at app.RUBY$block$\=app\,rb$1(app.rb:12)
	at java.lang.invoke.LambdaForm$DMH/93805087.invokeStatic_L7_L(LambdaForm$DMH)
	at java.lang.invoke.LambdaForm$MH/1489069835.invokeExact_MT(LambdaForm$MH)
	at org.jruby.runtime.CompiledIRBlockBody.commonYieldPath(CompiledIRBlockBody.java:70)
	at org.jruby.runtime.IRBlockBody.call(IRBlockBody.java:59)
	at org.jruby.runtime.Block.call(Block.java:106)
	at org.jruby.RubyProc.call(RubyProc.java:334)
	at org.jruby.RubyProc.call(RubyProc.java:240)
	at org.jruby.internal.runtime.RubyRunnable.run(RubyRunnable.java:99)
	at java.lang.Thread.run(Thread.java:745)
[nil]
made it past all the threads
```

ABaldwinHunter: Oh heh didn't refresh. Basically what @wfleming said :)

PR #134, opened 2016-09-30
Deploy channel branches automatically

wfleming:
This was a bit of a drive-by change in #132, but we want to keep this.

cc @codeclimate/review


maxjacobson:
LGTM

PR #135, opened 2016-10-06
Support python3

wfleming:
By default we use `python2` in this engine, which unfortunately chokes on some newer python syntax. This updates the engine to support both versions of Python: the default remains 2, but 3 can be configured with a setting in `.codeclimate.yml`:

```
engines:
  duplication:
    enabled: true
    config:
      languages:
        python:
          python_version: 3
```

The `parser.py` script has been updated to work with both versions of python, and the `parser.rb` has been updated to switch which python binary it will use depending on the config.

cc @codeclimate/review

### DIFF
--- a/lib/cc/engine/analyzers/python/main.rb
+++ b/lib/cc/engine/analyzers/python/main.rb
@@ -22,7 +22,11 @@ module CC
           end
 
           def parser(path)
-            ::CC::Engine::Analyzers::Python::Parser.new(self, File.binread(path), path)
+            ::CC::Engine::Analyzers::Python::Parser.new(python_version, File.binread(path), path)
+          end
+
+          def python_version
+            engine_config.languages.fetch("python", {}).fetch("python_version", 2)
           end
         end
       end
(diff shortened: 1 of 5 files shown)


lib/cc/engine/analyzers/python/main.rb
```diff
@@ -18,7 +18,15 @@ class Main < CC::Engine::Analyzers::Base
...
-            Node.new(::CC::Engine::Analyzers::Python::Parser.new(File.binread(path), path).parse.syntax_tree, path).format
+            Node.new(parser(path).parse.syntax_tree, path).format
+          end
+
+          def parser(path)
+            ::CC::Engine::Analyzers::Python::Parser.new(python_version, File.binread(path), path)
+          end
+
+          def python_version
+            engine_config.languages.fetch("python", {}).fetch("python_version", 2)
```
pbrisbin: Thoughts on hoisting `2` to `DEFAULT_PYTHON_VERSION`?

pbrisbin:
Awesome. I'm happy to see how relatively easy it was to make this engine multi-version.

lib/cc/engine/analyzers/python/main.rb
wfleming: Sure sure. Done.

Issue #136, opened 2016-10-09
NoMethodError on Python Duplication engine


kengz:
Shits on fire yo. The same code got an error from the duplication engine. See the original [CC build here](https://codeclimate.com/github/kengz/aiva/builds/69). If you see the commit for the build, that python file causing the error is not changed from previous build. The [project is here](https://github.com/kengz/aiva).

```shell
OpenJDK 64-Bit Server VM warning: Using the ParNew young collector with the Serial old collector is deprecated and will likely be removed in a future release
NoMethodError error occurred processing file ./lib/client.py: aborting.NoMethodError error occurred processing file ./lib/py/hello_rb.py: aborting.

NoMethodError: undefined method `fetch' for "":String
          python_version at /usr/src/app/lib/cc/engine/analyzers/python/main.rb:30
                  parser at /usr/src/app/lib/cc/engine/analyzers/python/main.rb:26
            process_file at /usr/src/app/lib/cc/engine/analyzers/python/main.rb:22
                     run at /usr/src/app/lib/cc/engine/analyzers/analyzer_base.rb:23
  block in process_files at /usr/src/app/lib/cc/engine/analyzers/reporter.rb:38
            block in run at /usr/src/app/lib/cc/engine/analyzers/file_thread_pool.rb:22
```

maxjacobson:
Hey @kengz, thanks for opening this issue. We did [recently change this engine](https://github.com/codeclimate/codeclimate-duplication/pull/135) to add configurable support for Python 3. We meant it to be backwards-compatible, but it looks like it wasn't for you. We'll dig into that further as soon as we can.

In the meantime, a short term fix for your repo is that you can change your `.codeclimate.yml` from

```yaml
  duplication:
    enabled: true
    config:
      languages:
        javascript:
        python:
        ruby:
```

To:

```yaml
  duplication:
    enabled: true
    config:
      languages:
        - javascript
        - python
        - ruby
```

I just [forked your repo](https://github.com/maxjacobson/aiva/commit/cf4d82e6006b5299ba87ba5bf4b1cfc1c36d0be6) and [verified that it doesn't error](https://codeclimate.com/github/maxjacobson/aiva/builds).

Apologies for the inconvenience.

kengz:
@maxjacobson thanks for the quick reply. Another issue is the PR webhook (status check) shows error even tho the build on cc's site is ok. This has been the case for the past few days too, will it be fixed together too?

[...]

PR #137, opened 2016-10-10
Use fetch_language to coerce python config

wfleming:
Addresses #136

The changes made in #135 didn't account for configuring the languages
with empty hash contents. With YAML parsing, something like

```
languages:
  python:
```

Gets parsed as `{ "languages" => { "python" => "" } }`, leading to a
`NoMethodError` when the Python class attempted to discover the
configured Python version.

`EngineConfig` already had logic to ensure a language configuration was
a hash, so this makes the appropriate coercing method public & uses it
from the Python class. (Since the Python version key is a Python-only
option, the `EngineConfig` class still doesn't seem like the appropriate
place for that logic.)

cc @codeclimate/review @maxjacobson

### DIFF
--- a/lib/cc/engine/analyzers/engine_config.rb
+++ b/lib/cc/engine/analyzers/engine_config.rb
@@ -30,10 +30,6 @@ module CC
           end
         end
 
-        private
-
-        attr_reader :config
-
         def fetch_language(language)
           language = config.
             fetch("languages", {}).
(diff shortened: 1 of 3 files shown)


maxjacobson:
Nice! Code change LGTM. Do you think it's worth adding a test?

wfleming:
Test added @maxjacobson. I'm merging.

PR #138, opened 2016-10-10
Raise visibility of Out Of Memory errors, and avoid them for minified JavaScript files

maxjacobson:
* allow OOM errors to become engine errors (unreverts an earlier PR)
* skip minified JS files (algorithm copied mostly from https://github.com/codeclimate/codeclimate-eslint/pull/115) because our reproduction case from an OSS repo was a large minified JS file.
* print paths to stderr before analysis, so if these errors continue, we'll be able to see which file led to the error (makes test output messy at the moment, unfortunately)

@codeclimate/review


lib/cc/engine/analyzers/file_thread_pool.rb
```diff
@@ -54,6 +54,13 @@ def thread_count
             DEFAULT_CONCURRENCY
           end
         end
+
+        def with_thread_abortion
```
pbrisbin: Cringe... How about, `#with_thread_abort_on_exceptions`?

lib/cc/engine/analyzers/analyzer_base.rb
```diff
@@ -20,7 +20,13 @@ def initialize(engine_config:)
         end
 
         def run(file)
-          process_file(file)
+          $stderr.puts("About to analyze #{file}. If analysis fails, consider excluding this file.")
```
pbrisbin: I would move this into the `else` so we don't show two messages in the skip case.

pbrisbin: I don't think we need to go so far as to say "If analysis fails", and this feels a little too wordy to me in general. Just saying `"Processing #{file}..."` is enough, IMO.

lib/cc/engine/analyzers/analyzer_base.rb
```diff
@@ -20,7 +20,13 @@ def initialize(engine_config:)
...
+          skip, skip_reason = skip?(file)
```
pbrisbin: To avoid the multiple return values, I think I'd prefer this method return a reason or `nil`, so it's just

```rb
if (reason = skip?(file)`
```

lib/cc/engine/analyzers/javascript/minification_checker.rb
```diff
@@ -0,0 +1,34 @@
...
+            line_count = 0
+            chars_seen = 0
+            content.chars.each do |char|
+              if char == newline
+                line_count += 1
+              else
+                chars_seen += 1
+              end
+            end
+            (chars_seen / line_count) >= MINIFIED_AVG_LINE_LENGTH_CUTOFF
```
pbrisbin: 💄 can you separate the return value with a blank line?

pbrisbin:
> makes test output messy at the moment, unfortunately

There's a bunch of ways we could quiet the specs, I think we should do that :)

maxjacobson:
@pbrisbin good calls! Updated

wfleming:
nit, but the PR title seems like the opposite of what we're doing: we're not avoiding the errors, we're increasing their visibility.

maxjacobson:
@wfleming true enough. The message only really describes the second commit. I'll update

lib/cc/engine/analyzers/javascript/minification_checker.rb
```diff
@@ -0,0 +1,26 @@
...
+        class MinificationChecker
+          MINIFIED_AVG_LINE_LENGTH_CUTOFF = 100
+
+          def initialize(path)
+            @content = File.read(path)
+          end
+
+          def minified?
+            line_count = content.lines.count
+            chars_seen = content.chars.count
```
pbrisbin: How about just

```rb
ratio = content.chars.count / content.lines.count
ratio >= MINIFIED_AVG_LINE_LENGTH_CUTOFF
```

(Note: I rescind my "white space before return value" suggestion in the case of a 2-line method).

lib/cc/engine/analyzers/file_thread_pool.rb
```diff
@@ -54,6 +54,13 @@ def thread_count
             DEFAULT_CONCURRENCY
           end
         end
+
+        def with_thread_abort_on_exceptions
+          t = Thread.new do
+            yield
+          end
+          (t.abort_on_exception = true) && t
```
pbrisbin: Our style guide allows the use of `and` to avoid these parenthesis. That said, I'd prefer this formatting:

```rb
def with_thread_abort_on_exceptions(&block)
  thread = Thread.new(&block)
  thread.abort_on_exception = true
  thread
end
```

FWIW

pbrisbin:
Some minor syntax suggestions, LGTM either way.

maxjacobson:
Resolved conflicts and going to merge on green

PR #139, opened 2016-10-11
Upgrade PhpParser to support php 7

maxjacobson:
This library has a handy upgrade guide here:
https://github.com/nikic/PHP-Parser/blob/master/UPGRADE-2.0.md

Seems there's just one backwards-incompatible change for our wrapper
code.

I'd like to get this into the next release before we release the OOM
fixes. Some php repos are using newer syntax and each file fails to
parse, and then it hits an OOM memory that users aren't seeing.

Example: https://codeclimate.com/github/etsy/phan/builds/277

Would like to verify on our phan fork on the beta channel that this will
work.

@codeclimate/review @ABaldwinHunter

### DIFF
--- a/spec/cc/engine/analyzers/php/main_spec.rb
+++ b/spec/cc/engine/analyzers/php/main_spec.rb
@@ -90,6 +90,17 @@ RSpec.describe CC::Engine::Analyzers::Php::Main, in_tmpdir: true do
         expect(run_engine(engine_conf)).to eq("")
       }.to output(/Skipping file/).to_stderr
     end
+
+    it "can parse php 7 code" do
+      create_source_file("foo.php", File.read(fixture_path("from_phan_php7.php")))
+      issues = run_engine(engine_conf).strip.split("\0")
+      result = issues.first.strip
+      json = JSON.parse(result)
+      expect(json["location"]).to eq({
+        "path" => "foo.php",
+        "lines" => { "begin" => 117, "end" => 118 },
+      })
+    end
   end
 
   def engine_conf
--- a/spec/fixtures/from_phan_php7.php
+++ b/spec/fixtures/from_phan_php7.php
@@ -1,0 +1,273 @@
+<?php declare(strict_types=1);
+namespace Phan;
+
+use ast\Node;
+
+/**
+ * Debug utilities
+ */
+class Debug
+{
+
+    /**
+     * Print a lil' something to the console to
+     * see if a thing is called
+     *
+     * @suppress PhanUnreferencedMethod
+     */
+    public static function mark()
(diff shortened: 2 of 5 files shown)


wfleming:
1. Is there any risk of PHP that parses successfully today failing to parse with this change?

1. I am surprised that multiple failures to parse would be related to OOM errors. The parsing is a separate process run for each file, so that memory should be recovered after parsing (or failing to parse). And if all the files fail, the main process isn't going to have many ASTs to hold in memory so that shouldn't increase much either. Can you explain more how these are related?

maxjacobson:
> Is there any risk of PHP that parses successfully today failing to parse with this change?

I want to do some QA to feel more confident about it, but I currently feel "pretty confident" about it, going based on

1. the test suite still passes, which exercises parsing some php 5 code
2. the changelog for 2.0 says this:

> PHP-Parser now requires PHP 5.4 or newer to run. It is however still possible to _parse_ PHP 5.2 and PHP 5.3 source code, while running on a newer version.

Re: your other question

> Can you explain more how these are related? 

Not really, no. Agreed it seems unrelated. Ashley pointed out shortly before I left the office that there were more OSS OOM error examples than I realized, and this one jumped out as having a lot of parsing error output and also an OOM error. It might be the case that even with the upgrade parser that phan will still have an OOM error and the parser errors were a red herring. I'm not sure yet.

maxjacobson:
Btw, I added a test exercising some php 7 code, which passes, and would fail without the upgrade (https://circleci.com/gh/codeclimate/codeclimate-duplication/545)

wfleming:
> Not really, no. Agreed it seems unrelated. Ashley pointed out shortly before I left the office that there were more OSS OOM error examples than I realized, and this one jumped out as having a lot of parsing error output and also an OOM error. It might be the case that even with the parser upgrade that phan will still have an OOM error and the parser errors were a red herring. I'm not sure yet.

My bet would be that enough files were still parsing successfully to grow memory usage and cause OOM errors. So this fix might make them happen faster 😄 .

wfleming:
Code LGTM.

ABaldwinHunter:
👏 

Curious about how our phan test repo runs on with this change.

maxjacobson:
OK we're taking bets now 😄 

Just pushed this change to the `channel/beta` branch, so we'll know soon.

maxjacobson:
@wfleming you win 😭 https://codeclimate.com/github/codeclimate-testing/phan/builds/3

maxjacobson:
I tested 2 more php 5 repos and they continue to be analyzed correctly, so I'm going to go ahead and ship this.

PR #142, opened 2016-10-18
add count_threshold option

### LINKED ISSUE #131: Only report if there are 3 occurrences of the code?
I've been fiddling with the mass settings on the duplication engine, and it's pretty good, but I find that often I have legitimate reasons for writing code twice; namely, it would be needlessly obfuscating code to refactor a given pair of functions.

Is there a way to configure the "count threshold"? Currently this seems to be hard coded at 2: if code happens twice, it is reported. I'd like to be able to configure something like this:

```
  duplication:
    enabled: true
    config:
      languages:
        ruby:
          mass_threshold: 10
          count_threshold: 3

```

This would emulate the "rule of three" in Ruby while being extra-strict about code mass.

If this is interesting to the core team, I could potentially even write a pull request.

devvmh:
adds the option to only check duplication if the code is found in more than just two locations.

fixes #131

### DIFF
--- a/lib/cc/engine/analyzers/engine_config.rb
+++ b/lib/cc/engine/analyzers/engine_config.rb
@@ -2,6 +2,8 @@ module CC
   module Engine
     module Analyzers
       class EngineConfig
+        DEFAULT_COUNT_THRESHOLD = 2
+
         def initialize(hash)
           @config = normalize(hash)
         end
@@ -31,8 +33,9 @@ module CC
         end
 
         def count_threshold_for(language)
-          threshold = fetch_language(language).fetch("count_threshold", nil)
-          threshold = config.fetch("count_threshold", nil) unless threshold
+          threshold = fetch_language(language)["count_threshold"] ||
+            config.fetch("config", {}).fetch("count_threshold", nil) ||
+            DEFAULT_COUNT_THRESHOLD
 
           if threshold
             threshold.to_i
--- a/lib/cc/engine/analyzers/engine_config.rb
+++ b/lib/cc/engine/analyzers/engine_config.rb
@@ -32,6 +32,7 @@ module CC
 
         def count_threshold_for(language)
           threshold = fetch_language(language).fetch("count_threshold", nil)
+          threshold = config.fetch("count_threshold", nil) unless threshold
 
           if threshold
             threshold.to_i
--- a/lib/cc/engine/analyzers/engine_config.rb
+++ b/lib/cc/engine/analyzers/engine_config.rb
@@ -30,6 +30,11 @@ module CC
           end
         end
 
+        def count_threshold_for(language)
+          threshold = fetch_language(language).fetch("count_threshold", nil)
+          threshold ? threshold.to_i : threshold
+        end
+
         def fetch_language(language)
           language = config.
             fetch("languages", {}).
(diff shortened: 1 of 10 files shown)


lib/cc/engine/analyzers/reporter.rb
```diff
@@ -51,6 +51,7 @@ def report
             violations = new_violations(issue)
 
             violations.each do |violation|
+              next if (violation.occurrences + 1) < language_strategy.count_threshold
```
devvmh: here's where the magic happens

lib/cc/engine/analyzers/engine_config.rb
```diff
@@ -30,6 +30,11 @@ def mass_threshold_for(language)
           end
         end
 
+        def count_threshold_for(language)
+          threshold = fetch_language(language).fetch("count_threshold", nil)
+          threshold ? threshold.to_i : threshold
```
dblandin: This ternary conditional is a bit difficult to read when the switch value is also the default return.

Given the similarity to the `#mass_threshold_for` method above, it might make sense to consolidate:
```ruby
threshold_config_for(:mass, "ruby")
threshold_config_for(:count, "ruby")

def threshold_config_for(option, language)
  threshold = fetch_language(language)["#{option}_threshold"]

  if threshold
   threshold.to_i
  end
end
```

devvmh:
I tried to update the spec, but it looks pretty involved because of where I put the code. Maybe I'll try again next week if I have time

dblandin:
@devvmh I'm good rolling with this as is. Going to do some local testing off of this branch before merging.

devvmh:
Great, thanks!

dblandin:
Hey @devvmh,

I've tested this locally and it works well! One question I have now that I've thought about it a bit more is whether this option should be configured one level up instead of nested under a particular language.

Do you think it's likely that users will want to specify different count thresholds per language? The `mass_threshold` configuration option works that way because of the difference in sensitivity between different language parser ASTs.

I'm thinking that it might be better to support this option as a sibling to `languages`:

```yaml
engines:
  duplication:
    enabled: true
    config:
      count_threshold: 3
      languages:
        - ruby
```

What do you think?

devvmh:
Yeah that's a good question. I think for my particular case I would do the same for ruby and javascript. I do think that it's more useful to be per-language than not though, don't you? Like, it's not harder to configure per-language (just 4 extra lines of YAML worst case).

If you'd like me to change it, just say and I can do that

dblandin:
Just took a quick poll internally, and the consensus was basically:

giphy

So I think it would be ideal if we supported the configuration level at the higher position, but also allowed language-specific overrides.

```yaml
engines:
  duplication:
    enabled: true
    config:
      count_threshold: 3
      languages:
        ruby:
        python:
        javascript:
          count_threshold: 5
```

Does that sound good to you?

devvmh:
Yeah! brilliant. I'll tty and implement that in tte jext few days

devvmh:
Phew, OK, that's finally updated.

dblandin:
@devvmh Awesome! The code looks great :+1: 

Last thing: Would you mind adding an example or two to [spec/cc/engine/analyzers/engine_config_spec.rb](https://github.com/codeclimate/codeclimate-duplication/blob/master/spec/cc/engine/analyzers/engine_config_spec.rb) to cover `EngineConfig#count_threshold_for`?

devvmh:
OK, updated. Should I squash this into one commit?

dblandin:
@devvmh Perfect :ok_hand: 

> Should I squash this into one commit?

No need. I'm going use GitHub's handy "Squash and merge" button.

Thanks for the contribution! :tada:

PR #144, opened 2016-10-24
[experiment, targets beta] Use spoon instead of open3

maxjacobson:
The idea here is that spoon uses the posix_spawn system call under the
hood instead of fork and exec. In theory this is more memory-efficient.

Some more context:
http://stackoverflow.com/questions/28286002/ruby-cannot-allocate-memory/28290263#28290263

Skipping over minified JS files seems like it will prevent most
customers from having problems, but we did see that etsy/phan still hits
OOM errors.

I'd like to try this on beta.


pbrisbin:
Is it fair to say we use the posix/spawn gem for these benefits whenever possible, but will use spoon here because posix/spawn doesn't support JRuby?

maxjacobson:
Yep, bingo: https://github.com/rtomayko/posix-spawn/issues/28#issuecomment-89399037

There are a few references to jruby in the posix-spawn codebase, but they seem to be vestigial. `bundle install` fails with posix-spawn

GordonDiggs:
Why didn't CC run on this PR?

pbrisbin:
@GordonDiggs because it's opened against a (non-default) channel branch.

GordonDiggs:
Ah right!

lib/cc/engine/analyzers/command_line_runner.rb
```diff
@@ -1,5 +1,6 @@
-require "open3"
 require "timeout"
+require 'spoon'
+require 'securerandom'
```
GordonDiggs: 💄 use double quotes

lib/cc/engine/analyzers/command_line_runner.rb
```diff
@@ -14,14 +15,33 @@ def initialize(command, timeout = DEFAULT_TIMEOUT)
 
         def run(input)
           Timeout.timeout(timeout) do
-            out, err, status = Open3.capture3(command, stdin_data: input)
+            file_actions = Spoon::FileActions.new
+            id = SecureRandom.hex
```
GordonDiggs: Does this have to be hex? We often use UUIDs

lib/cc/engine/analyzers/command_line_runner.rb
```diff
@@ -14,14 +15,33 @@ def initialize(command, timeout = DEFAULT_TIMEOUT)
...
 
-            if status.success?
-              yield out
+            # setup stdout
+            file_actions.close(1)
+            file_actions.open(1, "/tmp/#{id}.out", File::WRONLY | File::TRUNC | File::CREAT, 0600)
+
+            # setup stderr
+            file_actions.close(2)
+            file_actions.open(2, "/tmp/#{id}.err", File::WRONLY | File::TRUNC | File::CREAT, 0600)
```
maxjacobson: FWIW these options are the ones I'm least confident about. They're mostly copied from [this example](https://github.com/headius/spoon/blob/5b83982bf617834497d646fc21768e281ad1deda/examples/ls.rb).

The expression `File::WRONLY | File::TRUNC | File::CREAT` evaluates to `1537`. I'm not sure if `TRUNC` means that stderr and stdout will be truncated, but that wouldn't be great.

pbrisbin: I would guess it means, truncate the file when opening, like `>` vs `>>` on the commandline, but I'm not 100% sure.

lib/cc/engine/analyzers/command_line_runner.rb
```diff
@@ -30,26 +50,17 @@ def run(input)
...
-        # produced output that can be parsed as JSON.
-        #
-        # https://github.com/jruby/jruby/blob/master/lib/ruby/stdlib/open3.rb#L200-L201
-        #
-        def handle_open3_race_condition(out)
-          JSON.parse(out)
-          NullStatus.new(true, 0)
+        def successful_output?(id)
+          output = File.read("/tmp/#{id}.out")
+          JSON.parse(output, max_nesting: false) and output
```
GordonDiggs: if this is meant to return a boolean, this should use `&&`

lib/cc/engine/analyzers/command_line_runner.rb
```diff
@@ -30,26 +50,17 @@ def run(input)
...
         rescue JSON::ParserError
-          NullStatus.new(false, 1)
+          nil
```
GordonDiggs: since this is a `?` method, should this be `false`?

lib/cc/engine/analyzers/command_line_runner.rb
maxjacobson: This method is kind of a compound method, inspired by [this @pbrisbin comment](https://github.com/codeclimate/codeclimate-duplication/pull/138#discussion_r82686451), where it returns a useful truthy value when it's successful or a falsey value when it wasn't successful. In this case, I'm relying on `JSON.parse` to raise an error, so the `and` (or `&&`) is only used in the happy path, anyway. I'm tempted to change it to

```ruby
JSON.parse(output, max_nesting: false)
output
```

to make it more clear that we don't care about the truthiness of the first clause of the expression. WDYT?

lib/cc/engine/analyzers/command_line_runner.rb
wfleming: Pat's interpretation should be correct: the `TRUNC` flag is equivalent to [`O_TRUNC`](https://linux.die.net/man/2/open) under the hood, and is analogous to `>` in the shell, as he pointed out.

(Historical context: the reason for doing this if you're overwriting an existing file is that if the existing contents are, say, `ABCD`, and you open it *without* this flag, and write `12`, the final contents are `12CD`, which often isn't what you want.)

maxjacobson:
Thanks for the feedback. I'd like to go forward with shipping this to beta to see if it actually helps our remaining reproduction case.

maxjacobson:
> Thanks for the feedback. I'd like to go forward with shipping this to beta to see if it actually helps our remaining reproduction case.

It didn't: https://codeclimate.com/github/codeclimate-testing/phan/builds/4

Inclined to let this experiment die unless someone would like to bring this to master anyway.

PR #145, opened 2016-10-24
Guard against divide by zero error

ABaldwinHunter:
Fixes a bug where empty js file could cause an analysis to error

e.g. https://codeclimate.com/github/parliamentofowls/test-repo-cc/builds/1

@codeclimate/review

### DIFF
--- a/lib/cc/engine/analyzers/javascript/minification_checker.rb
+++ b/lib/cc/engine/analyzers/javascript/minification_checker.rb
@@ -10,8 +10,10 @@ module CC
           end
 
           def minified?
-            ratio = content.chars.count / content.lines.count
-            ratio >= MINIFIED_AVG_LINE_LENGTH_CUTOFF
+            if content.lines.count != 0
+              ratio = content.chars.count / content.lines.count
+              ratio >= MINIFIED_AVG_LINE_LENGTH_CUTOFF
+            end
           end
 
           private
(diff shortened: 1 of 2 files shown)


ABaldwinHunter:
I slightly disagree with the rubocop setting we have that's telling me to

```
Use content.lines.count.nonzero? instead of content.lines.count != 0.
```
because I think it's less explicit and mathematical, but going to comply for now.

lib/cc/engine/analyzers/javascript/minification_checker.rb
```diff
@@ -10,8 +10,10 @@ def initialize(path)
           end
 
           def minified?
-            ratio = content.chars.count / content.lines.count
-            ratio >= MINIFIED_AVG_LINE_LENGTH_CUTOFF
+            if content.lines.count != 0
+              ratio = content.chars.count / content.lines.count
+              ratio >= MINIFIED_AVG_LINE_LENGTH_CUTOFF
+            end
```
wfleming: I saw a comment on another PR this morning suggesting that a `?` method should probably return a bool, which made sense to me. In which case an `else... false` branch here might make sense.

ABaldwinHunter:
@wfleming updated!

wfleming:
LGTM

maxjacobson:
Oo good catch. This was inspired by the eslint checker y'all wrote a few months ago, but I forgot that JS is a little more loosey-goosey with math 😬

PR #146, opened 2016-10-26
Refactor away from OpenStruct for php AST

maxjacobson:
I believe this is the root cause for many of the OOM errors we've been
seeing.

Our original analysis pointed toward:

1. repos with big minified js files (resolved by skipping those)
1. php-only repos (mysterious...)

Since ripping off the bandaid Monday and surfacing OOM errors to
end-users, we've been hearing about several PHP-only repos that are
erroring, so I took some time to read through the php parsing code
looking for a possible memory leak.

Eventually I noticed that we're using OpenStruct to represent our "Node"
concept for the php AST that we build. I've heard that OpenStruct isn't
very memory-efficient or fast and decided to try refactoring away from
it.

I'm able to reproduce the OOM error via the CLI on my clone of
etsy/phan, and the error no longer occurs with this refactor.

:tada: :tada: :tada:

### DIFF
--- a/lib/cc/engine/analyzers/php/nodes.rb
+++ b/lib/cc/engine/analyzers/php/nodes.rb
@@ -5,8 +5,13 @@ module CC
     module Analyzers
       module Php
         module Nodes
-          class Node < OpenStruct
+          class Node
+            def initialize(attrs)
+              @attrs = attrs
+            end
+
             def accept(visitor)
+              node_type = @attrs[:node_type]
               visit_method = :"visit_#{node_type}Node"
 
               if visitor.respond_to?(visit_method)
@@ -15,11 +20,19 @@ module CC
             end
 
             def line
-              startLine
+              @attrs[:startLine]
+            end
+
+            def file
+              @attrs[:file]
             end
 
             def to_sexp
               CC::Engine::Analyzers::Php::SexpVisitor.new.accept(self)
+            end
+
+            def each_pair(&block)
+              @attrs.each_pair(&block)
             end
 
             def sub_nodes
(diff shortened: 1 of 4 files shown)


maxjacobson:
@codeclimate/review 👀 (feel free to update branch or merge for me if you like)

lib/cc/engine/analyzers/php/nodes.rb
```diff
@@ -5,7 +5,19 @@ module Engine
...
       module Php
         module Nodes
-          class Node < OpenStruct
+          class Node
+            def initialize(attrs)
+              @attrs = attrs
+            end
+
+            def method_missing(m, *args, &block)
+              if args.length.zero?
```
pbrisbin: WDYT of this:

```rb
def method_missing(key, *)
  @attrs.fetch(key) { super }
end

def respond_to_missing?(key)
  @attrs.key?(key)
end
```

ABaldwinHunter:
👏 LGTM

ABaldwinHunter:
Great find

lib/cc/engine/analyzers/php/nodes.rb
maxjacobson: oo missed this comment before pushing up a blunter implementation. Considering it offline now

maxjacobson:
@ABaldwinHunter thanks! Kind of a wild hunch that may have (_knock on wood_) worked out

@ABaldwinHunter @pbrisbin I pushed up one last tweak removing the `method_missing` stuff and squashed. Any more thoughts before I ship this?

pbrisbin:
Nice. LGTM.

PR #147, opened 2016-11-01
Double the stack memory limit

maxjacobson:
We're seeing some reports that analysis is failing with output like
this:

```
Error: Your application used more stack memory than the safety cap of 2048K.
Specify -J-Xss####k to increase it (#### = cap size in KB).
Specify -w for full StackOverflowError stack trace
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,6 +27,6 @@ RUN chown -R app .
 
 USER app
 
-ENV JAVA_OPTS="-XX:+UseParNewGC -XX:MinHeapFreeRatio=5 -XX:MaxHeapFreeRatio=10"
+ENV JAVA_OPTS="-XX:+UseParNewGC -XX:MinHeapFreeRatio=5 -XX:MaxHeapFreeRatio=10 -Xss4096k"
 
 CMD ["/usr/src/app/bin/duplication"]


maxjacobson:
@codeclimate/review WDYT about this? If the user's problem persists we'll need to dig deeper into the JavaScript AST building to see if there's any optimization to make there.

wfleming:
Seems reasonable for a stack limit.

Do you know if the default of this value is static or based on overall java memory limits? I.e. the default value of `Xmx` is determined based on machine memory, I'm wondering if this also is, or if it's always `2048K`.

maxjacobson:
@wfleming Looks like it's not based on machine memory, but based on platform: http://docs.oracle.com/cd/E13150_01/jrockit_jvm/jrockit/jrdocs/refman/optionX.html#wp1024112

maxjacobson:
So I think it will always be 2048 until we update our docker base image so it's running in a different context. Is that your reading of it?

wfleming:
2048 is already double what those docs say it should be, so that's weird.

Docker base image wouldn't affect it: you would have to actually be running a different host kernel. Nobody really uses 32bit anymore, so I'm not too worried about that.

So this LGTM.

maxjacobson:
Yea I wasn't sure about that... figured it wasn't up-to-date maybe? Going to go ahead and ship this though

PR #149, opened 2016-11-03
Disable debug output by default

maxjacobson:
This was helpful for identifying potentially problematic files, and
helped lead us to identify minified js files as a problem, but it hasn't
been as helpful recently, and can actually lead to exceeding the maximum
output limit for repos with many files!


maxjacobson:
@codeclimate/review 👀

dblandin:
It looks like this engine already supports a debug configuration option:

```yaml
engines:
  duplication:
    config:
      debug: true
```

Any reason not to guard with that instead of removing completely?

maxjacobson:
@dblandin oh, good catch! Didn't notice that. Updated

dblandin:
LGTM

PR #151, opened 2016-11-09
Allow hash-in-array configuration style

wfleming:
We saw some configs using this kind of style in the wild:

```
languages:
  - ruby:
    mass_threshold: 25
  - javascript
```

Right now, this config doesn't actually work (it gets eliminated due to validations in `cc-yaml`, and what the engine actually sees is `languages: ["javascript"]`, which is certainly not what the user wanted.

I'm working to change that behavior in `cc-yaml`, and wanted to also try to ensure that the duplication engine would have reasonable behavior once a fixed `cc-yaml` would make this style of config actually get passed through to the engine.

This adds support for this style of config, so it should start actually working once `cc-yaml` is fixed. It also explicitly raises an error for some degenerate cases where we would previously have just used the default configuration: I believe it's better to explicitly fail than to do something probably not desired by the user in those cases.

cc @codeclimate/review

### DIFF
--- a/lib/cc/engine/analyzers/engine_config.rb
+++ b/lib/cc/engine/analyzers/engine_config.rb
@@ -3,6 +3,7 @@ module CC
     module Analyzers
       class EngineConfig
         DEFAULT_COUNT_THRESHOLD = 2
+        InvalidConfigError = Class.new(StandardError)
 
         def initialize(hash)
           @config = normalize(hash)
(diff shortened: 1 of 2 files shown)


toddmohney:
I agree with raising an error on invalid key count. Fits well with the Zero, One, Infinity rule.
LGTM

PR #153, opened 2016-11-16
A few small things to help with debugging

maxjacobson:
Details in the commit messages

### DIFF
--- a/lib/cc/engine/analyzers/engine_config.rb
+++ b/lib/cc/engine/analyzers/engine_config.rb
@@ -22,7 +22,7 @@ module CC
         end
 
         def concurrency
-          config.fetch("concurrency", 2)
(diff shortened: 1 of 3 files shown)


GordonDiggs:
LGTM

lib/cc/engine/analyzers/engine_config.rb
```diff
@@ -22,7 +22,7 @@ def languages
         end
 
         def concurrency
-          config.fetch("concurrency", 2)
+          config.fetch("config", {}).fetch("concurrency", 2).to_i
```
toddmohney: Not sure if we're using a late enough version of ruby, but you could use the `dig` method with default fallbacks here.
http://ruby-doc.org/core-2.3.0_preview1/Hash.html#method-i-dig

Certainly, not a merge blocker. I just like the `:dig` interface :wink:

PR #154, opened 2016-11-17
Coerce debug config to boolean

maxjacobson:
I toggled "true" to "false" and it kept debugging. I should've just
removed the line, I guess, but I didn't want to.


wfleming:
Sure sure. LGTM

PR #157, opened 2017-01-24
Custom patterns

pointlessone:
Example config:
```yml
engines:
  duplication:
    enabled: true
    config:
      languages:
        foolang:
          patterns:
          - "*.foofile"
          - "Foofile"
```

### DIFF
--- a/lib/cc/engine/analyzers/engine_config.rb
+++ b/lib/cc/engine/analyzers/engine_config.rb
@@ -55,6 +55,10 @@ module CC
           end
         end
 
+        def patterns_for(language, fallbacks)
+          Array(fetch_language(language).fetch("patterns", fallbacks))
+        end
+
         private
 
         attr_reader :config
--- a/README.md
+++ b/README.md
@@ -105,6 +105,29 @@ engines:
     - examples/
 ```
 
+### Custom file name patterns
+
+All engines check only appropriate files but you can override default set of
+patterns. Patterns are ran aginast the project root direcory so you have to use
+`**` to match files in nested directories. Also note that you have to specify
+all patterns, not only the one you want to add.
+
+```yml
+engines:
+  duplication:
+    enabled: true
+    config:
+      languages:
+        ruby:
+          patterns:
+            - "**/*.rb
+            - "**/*.rake"
+            - "Rakefile"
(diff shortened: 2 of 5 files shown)


lib/cc/engine/analyzers/engine_config.rb
```diff
@@ -55,6 +55,10 @@ def fetch_language(language)
           end
         end
 
+        def patterns_for(language, fallbacks)
+          Array(fetch_language(language).fetch("patterns", fallbacks))
```
wfleming: Nice simple way of ensuring you get the type you want, cool. 👍

lib/cc/engine/analyzers/engine_config.rb
```diff
@@ -55,6 +55,10 @@ def fetch_language(language)
           end
         end
 
+        def patterns_for(language, fallbacks)
```
wfleming: The integration-y test you have is great. Would it be reasonable to get some unit tests as well? For example, in addition to the happy path, it'd be nice to have a test that a malformed config like:

```
ruby:
  patterns: "*.rb"
```

does the right thing. (I think your code does handle this quite nicely, but guarding against future regressions with a test would be nice.)

lib/cc/engine/analyzers/php/parser.rb
```diff
@@ -16,7 +16,7 @@ def initialize(code, filename)
           end
 
           def parse
-            runner = CommandLineRunner.new("php #{parser_path}")
+            runner = CommandLineRunner.new("php -d 'display_errors = Off' #{parser_path}")
```
wfleming: This looks like a good bugfix in its own right, and not directly related to the new feature. It's already in its own commit, so I'd support cherry-picking this into a new branch & shipping it separately as its own PR if you want to. 

Also, it's been many years since I wrote much PHP, so pardon if this is a silly question: this won't stop us from seeing exceptions/stack traces in logs if the parser crashes, will it?

pointlessone: PHP duplication specs were failing for me because notices polluted the output. This is definitely setup issue but I don't know if CI has similar setup to that of mine. I can extract this commit and see how CI will react to that.

PR #158, opened 2017-01-26
Suppress PHP notices

pointlessone:
PHP notices pollute output and trip over some specs.

Note: On CI PHP does not produce any notices. It happens when `rake spec` ran locally. This is most definitely a setup issue. I'm not sure how easy it's to mitigate it. On macOS `brew`'s PHP 5.6 spews a lot of notices in its default configuration. It makes development a bit challenging.

### DIFF
--- a/lib/cc/engine/analyzers/php/parser.rb
+++ b/lib/cc/engine/analyzers/php/parser.rb
@@ -16,7 +16,7 @@ module CC
           end
 
           def parse
-            runner = CommandLineRunner.new("php #{parser_path}")
+            runner = CommandLineRunner.new("php -d 'display_errors = Off' #{parser_path}")
             runner.run(code) do |output|
               json = parse_json(output)
 


maxjacobson:
WDYT about using `make test` instead of `rake spec`?

pointlessone:
This is a Ruby project. Since this is not mentioned in the README I assumed it follows community convention to run tests with `rake spec` (well, rather `rake` default task). Very well might be I'm "[holding it wrong](https://www.engadget.com/2010/06/24/apple-responds-over-iphone-4-reception-issues-youre-holding-th/)". In that case I'd rather submit a README PR.

maxjacobson:
@pointlessone heh, nice reference

IMO tests must pass inside of docker (how `make test` runs them) and it's very nice if they pass outside of docker, so I'm sympathetic to this proposed change, but slightly worried that we'll suppress actually useful information in production. This is what Will was asking about in #157:

> Also, it's been many years since I wrote much PHP, so pardon if this is a silly question: this won't stop us from seeing exceptions/stack traces in logs if the parser crashes, will it?

I'm also not sure of the answer to that, which is why I'm hesitant to approve this change. Do you have any additional context on that?

pointlessone:
This option affects only vendored PHP parser. It is not a development focus of this project, is it? So do we care much about the code we don't maintain?

Exceptions are going into stderr, not sdout. Notices, on the other hand, go into stdout by default. It's possible to redirect them to stderr. I can do that instead of completely disabling errors. This option does not disable exceptions in stderr.

So overall it doesn't look like we're going to lose anything with this change.

Moreover, this option should be `off` for production envs. Here's an excerpt from PHP documentation:

> This is a feature to support your development and should never be used on production systems (e.g. systems connected to the internet).

maxjacobson:
Ah, as long as exceptions still go to stderr and aren't suppressed, then this sounds good to me.

PR #161, opened 2017-02-18
Switch to MRI for flay

zenspider:
I think I got the image working, but I haven't tested the other engines at all. Testing just the ruby side there is a very clear benefit to using MRI.

This is mostly here at this point so others can poke at it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM jruby:9.1.5-jdk
+FROM ruby:2.3-alpine
 
 WORKDIR /usr/src/app/
 
@@ -10,23 +10,27 @@ COPY vendor/php-parser/composer.lock /usr/src/app/vendor/php-parser/
 
 COPY package.json /usr/src/app/
 
-RUN curl --silent --location https://deb.nodesource.com/setup_5.x | bash - && \
-    apt-get update && apt-get install -y nodejs python python3 openssh-client php5-cli php5-json
+# not sure if this is ok to skip. apk's nodejs is 6.7.x:
+# RUN curl --silent --location https://deb.nodesource.com/setup_5.x | bash -
+RUN apk update && apk add nodejs python python3 php5-phar php5-openssl php5-cli php5-json php5-zlib
+
+# for bundler w/ flay fork... not necessary once off flay fork
+RUN apk add git
 RUN gem install bundler --no-ri --no-rdoc && \
-    bundle install -j 4 && \
-    curl -sS https://getcomposer.org/installer | php
+    bundle install -j 4
+
+RUN apk add curl
+RUN curl -sS https://getcomposer.org/installer | php
+RUN apk del --purge curl git
 
 RUN mv composer.phar /usr/local/bin/composer
 RUN cd /usr/src/app/vendor/php-parser/ && composer install --prefer-source --no-interaction
 RUN npm install
 
-RUN adduser app -u 9000
-
+RUN adduser -u 9000 -D -h /usr/src/app -s /bin/false app
 COPY . /usr/src/app
-RUN chown -R app .
+RUN chown -R app:app /usr/src/app
 
 USER app
-
-ENV JAVA_OPTS="-XX:+UseParNewGC -XX:MinHeapFreeRatio=5 -XX:MaxHeapFreeRatio=10 -Xss4096k"
 
 CMD ["/usr/src/app/bin/duplication"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,11 +10,8 @@ COPY vendor/php-parser/composer.lock /usr/src/app/vendor/php-parser/
 
 COPY package.json /usr/src/app/
 
-# not sure if this is ok to skip. apk's nodejs is 6.7.x:
-# RUN curl --silent --location https://deb.nodesource.com/setup_5.x | bash -
 RUN apk update && apk add nodejs python python3 php5-phar php5-openssl php5-cli php5-json php5-zlib php5-xml
 
-# git is for bundler w/ flay fork... not necessary once off flay fork
 RUN apk add curl && \
     gem install bundler --no-ri --no-rdoc && \
     bundle install -j 4 && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,14 +14,12 @@ COPY package.json /usr/src/app/
 # RUN curl --silent --location https://deb.nodesource.com/setup_5.x | bash -
 RUN apk update && apk add nodejs python python3 php5-phar php5-openssl php5-cli php5-json php5-zlib
 
-# for bundler w/ flay fork... not necessary once off flay fork
-RUN apk add git
-RUN gem install bundler --no-ri --no-rdoc && \
-    bundle install -j 4
-
-RUN apk add curl
-RUN curl -sS https://getcomposer.org/installer | php
-RUN apk del --purge curl git
+# git is for bundler w/ flay fork... not necessary once off flay fork
+RUN apk add git curl && \
+    gem install bundler --no-ri --no-rdoc && \
+    bundle install -j 4 && \
+    curl -sS https://getcomposer.org/installer | php && \
+    apk del --purge curl git
 
 RUN mv composer.phar /usr/local/bin/composer
 RUN cd /usr/src/app/vendor/php-parser/ && composer install --prefer-source --no-interaction
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,14 +12,14 @@ COPY package.json /usr/src/app/
 
 # not sure if this is ok to skip. apk's nodejs is 6.7.x:
 # RUN curl --silent --location https://deb.nodesource.com/setup_5.x | bash -
-RUN apk update && apk add nodejs python python3 php5-phar php5-openssl php5-cli php5-json php5-zlib
+RUN apk update && apk add nodejs python python3 php5-phar php5-openssl php5-cli php5-json php5-zlib php5-xml
 
 # git is for bundler w/ flay fork... not necessary once off flay fork
-RUN apk add git curl && \
+RUN apk add curl && \
     gem install bundler --no-ri --no-rdoc && \
     bundle install -j 4 && \
     curl -sS https://getcomposer.org/installer | php && \
-    apk del --purge curl git
+    apk del --purge curl
 
 RUN mv composer.phar /usr/local/bin/composer
 RUN cd /usr/src/app/vendor/php-parser/ && composer install --prefer-source --no-interaction
--- a/lib/cc/engine/analyzers/ruby/main.rb
+++ b/lib/cc/engine/analyzers/ruby/main.rb
@@ -26,8 +26,6 @@ module CC
             RubyParser.new.process(File.binread(file), file, TIMEOUT)
           rescue Timeout::Error
             warn "TIMEOUT parsing #{file}. Skipping."
-          rescue => e
-            warn "ERROR parsing #{file}: #{e.class}: #{e.message}. Skipping"
           end
         end
       end
--- a/lib/cc/engine/analyzers/ruby/main.rb
+++ b/lib/cc/engine/analyzers/ruby/main.rb
@@ -24,6 +24,10 @@ module CC
 
           def process_file(file)
             RubyParser.new.process(File.binread(file), file, TIMEOUT)
+          rescue Timeout::Error
+            warn "TIMEOUT parsing #{file}. Skipping."
+          rescue => e
+            warn "ERROR parsing #{file}: #{e.class}: #{e.message}. Skipping"
           end
         end
       end
--- a/lib/ccflay.rb
+++ b/lib/ccflay.rb
@@ -6,15 +6,15 @@ require "digest"
 # A thread-safe and stable hash subclass of Flay.
 
 class CCFlay < Flay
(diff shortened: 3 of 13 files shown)


Dockerfile
```diff
@@ -10,23 +10,27 @@ COPY vendor/php-parser/composer.lock /usr/src/app/vendor/php-parser/
...
 COPY package.json /usr/src/app/
 
-RUN curl --silent --location https://deb.nodesource.com/setup_5.x | bash - && \
-    apt-get update && apt-get install -y nodejs python python3 openssh-client php5-cli php5-json
+# not sure if this is ok to skip. apk's nodejs is 6.7.x:
+# RUN curl --silent --location https://deb.nodesource.com/setup_5.x | bash -
+RUN apk update && apk add nodejs python python3 php5-phar php5-openssl php5-cli php5-json php5-zlib
+
+# for bundler w/ flay fork... not necessary once off flay fork
+RUN apk add git
```
GordonDiggs: Though it's a bit counterintuitive, it's better to collapse layers in this file when you are deleting things at the end. It helps keep image size down because the size is determined by the size at the end of the layer's command. In this case, I would collapse it to

```
RUN apk add git curl && \
    gem install bundler --no-ri --no-rdoc && \
    bundle install -j 4 && \
    curl -sS https://getcomposer.org/installer | php && \
    apk del --purge curl git
```

Rakefile
```diff
@@ -5,3 +5,49 @@ begin
   task :default => :spec
 rescue LoadError
 end
+
+ENV["CODECLIMATE_DEBUG"] = "1" if ENV["DEBUG"]
+
+IMAGE_NAME = "codeclimate/codeclimate-duplication"
+
+task :image do
+  sh "docker build --rm -t #{IMAGE_NAME} ."
```
GordonDiggs: We prefer to use `make` for these - this repo already has a `Makefile`. Can you move tasks there if they don't already exist?

zenspider: This is just my junk to get throughout the day. I can remove it from the PR...

GordonDiggs:
Did some testing and this is looking good functionality-wise! Results are the same across languages for the new and old engines

Dockerfile
```diff
@@ -10,23 +10,25 @@ COPY vendor/php-parser/composer.lock /usr/src/app/vendor/php-parser/
 
 COPY package.json /usr/src/app/
 
-RUN curl --silent --location https://deb.nodesource.com/setup_5.x | bash - && \
-    apt-get update && apt-get install -y nodejs python python3 openssh-client php5-cli php5-json
-RUN gem install bundler --no-ri --no-rdoc && \
+# not sure if this is ok to skip. apk's nodejs is 6.7.x:
+# RUN curl --silent --location https://deb.nodesource.com/setup_5.x | bash -
```
GordonDiggs: What you have should be fine - can you remove these comments?

Dockerfile
```diff
@@ -10,23 +10,25 @@ COPY vendor/php-parser/composer.lock /usr/src/app/vendor/php-parser/
...
+RUN apk update && apk add nodejs python python3 php5-phar php5-openssl php5-cli php5-json php5-zlib php5-xml
+
+# git is for bundler w/ flay fork... not necessary once off flay fork
```
GordonDiggs: Remove this comment too

lib/ccflay.rb
```diff
@@ -0,0 +1,86 @@
...
+
+    self.identical = Concurrent::Hash.new
+    self.masses = Concurrent::Hash.new
+  end
+
+  ##
+  # Calculate the structural hash for this sexp. Cached, so don't
+  # modify the sexp afterwards and expect it to be correct.
+
+  if ENV["PURE_HASH"]
```
GordonDiggs: Do we need to have both of these implementations?

spec/spec_helper.rb
```diff
@@ -33,11 +33,11 @@ def method_missing(*)
 
   config.before(:each) do
     $stderr = DummyStderr.new
-  end
+  end unless ENV["ENGINE_DEBUG"]
```
GordonDiggs: Can you wrap this whole set in this check?

```ruby
unless ENV["ENGINE_DEBUG"]
  config.before(:each) do
  ...

  config.after(:each) do
  ...
end
```

lib/ccflay.rb
zenspider: No, but it'll allow us to A/B test them w/ a very minor ENV change. The pure version seems to be much nicer to memory footprint at the least. Benchmarks faster but that doesn't seem to affect wall clock on real projects (not the bottleneck on bigger projects).

Dockerfile
zenspider: done

Dockerfile
zenspider: done

spec/spec_helper.rb
zenspider: done

lib/cc/engine/analyzers/ruby/main.rb
```diff
@@ -24,6 +24,10 @@ class Main < CC::Engine::Analyzers::Base
 
           def process_file(file)
             RubyParser.new.process(File.binread(file), file, TIMEOUT)
+          rescue Timeout::Error
```
wfleming: I'm in favor of rescuing these (I thought we already did: maybe we used to & it got removed?), but if we rescue it for ruby we should probably rescue it for all languages. So this rescue is probably better off in `AnalyzerBase#run`?

zenspider: I rescue it for ruby because RubyParser specifically raises it. If the other languages don't raise, then it shouldn't be pushed up.

I don't actually like `AnalyzerBase::RESCUABLE_ERRORS`... it's problematic in that you need to handle each and every type of exception that a given engine might raise. That's breaking responsibilities and should be pushed down. If necessary, a given engine should wrap an exception and raise an engine specific exception to be handled generically.

lib/cc/engine/analyzers/ruby/main.rb
```diff
@@ -24,6 +24,10 @@ class Main < CC::Engine::Analyzers::Base
...
+            warn "TIMEOUT parsing #{file}. Skipping."
+          rescue => e
```
zenspider: Yup. I don't like it. In the case of ruby parser, there was a case that ran fine in jruby (probably a bug, if not just a semantic difference) but raised in MRI. It was a RegexpError and not in the list. If RubyParser raises an exception, it is almost invariably because of something like this. Logging it to stderr helped me _find_ the bug, not hide it. For whatever reason, the way things currently worked the whole thing broke down and none of stderr was printed. This fixes it so I can actually get info and fix bugs further down.

zenspider: Gah. I hadn't actually paid attention to the code doing the rescuing in `#run`... Wouldn't this make more sense?

```ruby
        rescue => ex
          case ex
          when *RESCUABLE_ERRORS then
            $stderr.puts("Skipping file #{file} due to exception (#{ex.class}): #{ex.message}\n#{ex.backtrace.join("\n")}")
          else
            $stderr.puts("#{ex.class} error occurred processing file #{file}: aborting.")
            raise ex
          end
```

lib/cc/engine/analyzers/ruby/main.rb
wfleming: > if the other languages don't raise, then it shouldn't be pushed up.

I believe they do, though: the other languages use `CommandLineRunner`, which uses `Timeout`, and I don't think that's already caught by those.

lib/cc/engine/analyzers/ruby/main.rb
wfleming: > Wouldn't this make more sense?

That's actually pretty similar what the code used to be! But there's a good reason we don't do it that way any more.

The reason we don't use `case ... when *RESCUABLE_ERRORS` (or `rescue *RESCUABLE_ERRORS =>`, which has very similar semantics), is that when that collection includes class `A`, and the raised exception is of class `B < A`, it matches. Which seems sensible at first brush.

Unfortunately, you'll note that we rescue `::RuntimeError`. We do this because there are some bare `raise`s we hit on parser issues (I think somewhere in `RubyParser`, or possibly even in `Racc`?) that we wanted to skip, but we felt rescuing from all potential children of `RuntimeError` was excessive. So we have the logic you see today which only matches exact instances of rescuable exception classes (though even the current logic for that could probably be cleaned up without changing the behavior).

> It was a RegexpError and not in the list
> I don't actually like AnalyzerBase::RESCUABLE_ERRORS...That's breaking responsibilities and should be pushed down

I agree with you, and I'd be happy to see a PR break that down so that list is per-analyzer-class, but since this PR is focused on MRI, maybe that should be a separate PR? Same goes for the timeout handling, actually: AFAICT we weren't rescuing those before on JRuby, so that also would make sense to me as a separate changeset.

So maybe:

1. Back out the two rescues here so this is more focused on getting us on MRI
2. Open another PR to:
    - Rescue timeouts globally (since I believe all analyzers do raise them, correct me if I missed something)
   - Break out rescuable exception classes per-analyzer-class
   - ... and add `RegexpError` to the errors the ruby parser will catch.

Would that make sense?

lib/cc/engine/analyzers/ruby/main.rb
zenspider: Ah! True... any reason why `CommandLineRunner` shouldn't rescue the timeout and raise `CC::Engine::Analyzers::ParserError`?

lib/cc/engine/analyzers/ruby/main.rb
zenspider: Is there some reason why is a `map` + `include`? instead of `find` or `any?`?

zenspider: 1. Those rescues were actually necessary to get the code running on MRI, so I'd like to leave them.
2. Given permission from @GordonDiggs... I'll do the followup PR to clean this up and get error handling generally better. Otherwise I'll open an issue pointing here so we have an artifact to track.

zenspider:
```
          if RESCUABLE_ERRORS.include? ex.class
# or
          if RESCUABLE_ERRORS.any? { |klass| ex.instance_of?(klass) }
```

GordonDiggs: > Those rescues were actually necessary to get the code running on MRI, so I'd like to leave them.

What was erroring without them? Everything or only 2.4 code?

> I'll do the followup PR to clean this up and get error handling generally better. Otherwise I'll open an issue pointing here so we have an artifact to track.

I'm cool with this idea if it let's us remove the bare rescues shortly. I would hold off on deploying this image after merging the PR, though. Suppressing general exceptions in engines in production has been problematic before

lib/cc/engine/analyzers/ruby/main.rb
wfleming: Semantically, I think a timeout is distinct from a `ParserError`. A parse error (hopefully) indicates invalid syntax (though of course in practice with external parsers it  could also mean a parser bug leading to a crash). A timeout could indicate that the file was just *really* big/complicated, or could also indicate a parser bug.

I think it's valuable to raise/log more specific errors types when available. In a better world, it'd be nice to have more specific sub-classes of `ParserError`, but that's difficult due to the nature of these parsers being external processes. But timeout is a case that we can keep distinct, so I think we should.

lib/cc/engine/analyzers/ruby/main.rb
zenspider: I don't remember the specifics anymore. It's been a while and I'm sleep deprived today. I can roll out that rescue and rerun my gauntlet if necessary.

Otherwise, I guess objections have been met and this should go in? Then I can clean up error handling.

lib/cc/engine/analyzers/ruby/main.rb
zenspider: Not sure I entirely agree but this is arguing semantics at this point. `CC::Engine::Analyzers::ParserError` means anything we want. It can certainly mean "Syntax errors only", but by name, it means there was an error in the parser. Whether that is "syntax error", "took too long", "not supported", or whatever... The exception can be made to wrap another exception to provide more details if we want, tho, in practice, I've found that the CC engine generally does NOT provide more details where there are errors. That's why I rescued and logged where I did.

lib/cc/engine/analyzers/ruby/main.rb
wfleming: > Otherwise, I guess objections have been met and this should go in?

Sorry, I may have might be slightly unclear saying "fine with me to keep these in this PR...", or I'm misunderstanding what you're saying here 😃. I am not comfortable shipping a bare rescue that swallows all exceptions it catches as warnings.

What I meant was: if the change to rescue behavior is necessary to avoid an overall behavior change  (i.e. there is code that was parsed successfully on JRuby but not the new MRI) branch, then hashing out the right approach to the exception handling in this PR is fine. That applies to both this question around the RegExp errors and the timeout handling.

GordonDiggs:
@wfleming @zenspider What's left to ship this PR?

zenspider:
@GordonDiggs I need to clean up one rescue still... but I got distracted and did this first: https://github.com/codeclimate/codeclimate/pull/582

PR #164, opened 2017-03-16
Added Sexp.flatter and use that to optimize non-ruby analysis.

zenspider:
Specifically, for a fairly normal sized javascript file, the parsed
and converted sexp had a mass of 44381 and a depth of 191 (and in
would get a "stack level too deep" error calculating the depth). This
is because every key/value pair in the json became 2 levels of tree
when converted to a sexp.

Now, calling `sexp.flatter` before passing to flay converts that same
tree to have a mass of 22965 and a depth of 91, all while being
isomorphic to the original.

This takes the time to process that file from 27 seconds to 7. And the
project I was benchmarking went from timing out at 10 minutes to
finishing in 3.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -66,9 +66,6 @@ RUN adduser -u 9000 -D -h /usr/src/app -s /bin/false app
 COPY . /usr/src/app
 RUN chown -R app:app /usr/src/app
 
-# ENV PURE_HASH=1
-# ENV CODECLIMATE_PROFILE=1
-
 USER app
 
 CMD ["/usr/src/app/bin/duplication"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,9 +10,40 @@ COPY vendor/php-parser/composer.lock /usr/src/app/vendor/php-parser/
 
 COPY package.json /usr/src/app/
 
-RUN apk update && apk add nodejs python python3 php5-phar php5-openssl php5-cli php5-json php5-zlib php5-xml
+RUN apk update && apk add python python3 php5-phar php5-openssl php5-cli php5-json php5-zlib php5-xml
 
-RUN apk add curl && \
+ENV NODE_VERSION=v5.12.0 \
+    NPM_VERSION=3
+
+# Based on https://github.com/mhart/alpine-node
+RUN apk add --no-cache curl make gcc g++ linux-headers binutils-gold gnupg libstdc++ && \
+    gpg --keyserver ha.pool.sks-keyservers.net --recv-keys \
+      94AE36675C464D64BAFA68DD7434390BDBE9B9C5 \
+      FD3A5288F042B6850C66B31F09FE44734EB7990E \
+      71DCFD284A79C3B38668286BC97EC7A07EDE3FC1 \
+      DD8F2338BAE7501E3DD5AC78C273792F7D83545D \
+      C4F0DFFF4E8C1A8236409D08E73BC641CC11F4C8 \
+      B9AE9905FFD7803F25714661B63B535A4C206CA9 \
+      56730D5401028683275BD23C23EFEFE93C4CFFFE && \
+    curl -sSLO https://nodejs.org/dist/${NODE_VERSION}/node-${NODE_VERSION}.tar.xz && \
+    curl -sSL https://nodejs.org/dist/${NODE_VERSION}/SHASUMS256.txt.asc | gpg --batch --decrypt | \
+      grep " node-${NODE_VERSION}.tar.xz\$" | sha256sum -c | grep . && \
+    tar -xf node-${NODE_VERSION}.tar.xz && \
+    cd node-${NODE_VERSION} && \
+    ./configure --prefix=/usr ${CONFIG_FLAGS} && \
+    make -j$(getconf _NPROCESSORS_ONLN) && \
+    make install && \
+    cd / && \
+    if [ -x /usr/bin/npm ]; then \
+      npm install -g npm@${NPM_VERSION} && \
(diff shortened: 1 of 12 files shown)


maxjacobson:
⚠️ just want to mention, we shouldn't merge this until we fix the bug on master. Or we should revert  https://github.com/codeclimate/codeclimate-duplication/pull/161, which isn't currently deployed. Some context in case you missed it: ([slack1](https://codeclimate.slack.com/archives/C0306TRT2/p1489164362710176)) ([slack2](https://codeclimate.slack.com/archives/C0306TRT2/p1489176461678547)). Not sure who's owning fixing that, but want to make sure we don't accidentally re-ship it.

zenspider:
@wfleming the actual structure is isomorphic... the mass is changing because `s().mass == 1` and those are getting folded. The current default threshold for javascript is 40 (way more than ruby's 16) which might be high, but I don't have enough data on real world javascript to generate meaningful statistics... It shouldn't matter _too much_ regardless, because that threshold is just the farthest down we recurse, but all those structures are still getting analyzed via their parent nodes... so structural similarities of mass 40+ will still get reported properly... and that hasn't shifted that much given the `flatter` call...

Happy to change the default if someone wants to run some stats (or come up with something arbitrary... I don't have a vested interest in a number).

zenspider:
@maxjacobson this _is_ addressing #161, at least partially... on the javascript side at least, some/most of the errors were timeouts. I don't have good data for PHP (don't know how popular it is on your service)

Dockerfile
```diff
@@ -10,13 +10,53 @@ COPY vendor/php-parser/composer.lock /usr/src/app/vendor/php-parser/
 
 COPY package.json /usr/src/app/
```
GordonDiggs: Would it be possible to move copy all of the lock files and package files below the update and install steps? Otherwise, bumping gems will invalidate all of these layers

Dockerfile
```diff
@@ -26,6 +66,9 @@ RUN adduser -u 9000 -D -h /usr/src/app -s /bin/false app
 COPY . /usr/src/app
 RUN chown -R app:app /usr/src/app
 
+# ENV PURE_HASH=1
```
GordonDiggs: We should either uncomment or remove these

lib/ccflay.rb
```diff
@@ -84,3 +84,25 @@ def pure_ruby_hash # :nodoc: see above
     hash
   end
 end
+
+class Sexp
+  def flatter
+    r = dup.clear
```
GordonDiggs: These variable names aren't very clear. Can you expand them to be more descriptive?

spec/cc/engine/analyzers/python/main_spec.rb
```diff
@@ -48,19 +48,19 @@
...
       })
-      expect(json["remediation_points"]).to eq(1_600_000)
+      expect(json["remediation_points"]).to eq(1_550_000)
       expect(json["other_locations"]).to eq([
         {"path" => "foo.py", "lines" => { "begin" => 2, "end" => 2} },
         {"path" => "foo.py", "lines" => { "begin" => 3, "end" => 3} },
       ])
-      expect(json["content"]["body"]).to match /This issue has a mass of 6/
-      expect(json["fingerprint"]).to eq("019118ceed60bf40b35aad581aae1b02")
+      expect(json["content"]["body"]).to match(/This issue has a mass of 5/)
```
GordonDiggs: What caused these masses to change?

spec/cc/engine/analyzers/python/main_spec.rb
```diff
@@ -48,19 +48,19 @@
...
+      expect(json["fingerprint"]).to eq("8941b71bb75571fca80cca37a3d23dc1")
```
GordonDiggs: Same question for these fingerprints - we need to be especially careful about changing fingerprints

wfleming: Fingerprints include mass, so I believe this is a result of the masses changing. I tried to cover what I think is the customer impact of that a bit here, though I was more preoccupied with the mass thresholds changing what's considered an issue: https://github.com/codeclimate/codeclimate-duplication/pull/164#pullrequestreview-27497512.

wfleming:
> so structural similarities of mass 40+ will still get reported properly... and that hasn't shifted that much given the flatter call...

The trouble is that *what nodes* have a mass of 40+ will change. Instances of duplication that would be reported as issues by CC today would no longer be reported as issues, and effectively invalidate existing customer configurations.

I constructed a concrete example JS file, inline below demonstrating this: With default settings on `master` today, the engine reports these two functions as similar and a mass of 42. Run using this branch, the engine finds no issues with the default config, but if I change the config to a lower mass threshold, it once again reports these two functions as similar with mass of 38.

Similar to Gordon's comment about needing to be careful about changing fingerprints, we need to be careful about changing the actual results users will receive.

```foo.js
function a() {
  var a = 0, b = 1;
  console.log("foo");
  if (1 + 1 - 5) {
    console.log("foo");
    console.log("foo");
  }
}

function b() {
  var a = 0, b = 1;
  console.log("foo");
  if (1 + 1 - 5) {
    console.log("foo");
    console.log("foo");
  }
}
```

Dockerfile
zenspider: The way the Dockerfile is currently written/architected, no. Bundle is part of the big build in order to get the package deletions to work. 

We can if we switch to a bunch of `RUN` lines and use that `--squash` command line option... but that needs experienced eyeballs on it.

GordonDiggs: Bummer. I might take a swing at that another time then

GordonDiggs:
Closing this in favor of #166 and #167

PR #166, opened 2017-03-20
Fix non-ruby speed problems

zenspider:
This fixes non-ruby speed problems caused by overzealous AST conversion into sexps. Depth of one project went from 191 to 91 and speed up from ~27 seconds to 6-7 seconds. That project now goes from timing out to finishing in 2-3 minutes.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -66,9 +66,6 @@ RUN adduser -u 9000 -D -h /usr/src/app -s /bin/false app
 COPY . /usr/src/app
 RUN chown -R app:app /usr/src/app
 
-# ENV PURE_HASH=1
-# ENV CODECLIMATE_PROFILE=1
-
 USER app
 
 CMD ["/usr/src/app/bin/duplication"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,9 +10,40 @@ COPY vendor/php-parser/composer.lock /usr/src/app/vendor/php-parser/
 
 COPY package.json /usr/src/app/
 
-RUN apk update && apk add nodejs python python3 php5-phar php5-openssl php5-cli php5-json php5-zlib php5-xml
+RUN apk update && apk add python python3 php5-phar php5-openssl php5-cli php5-json php5-zlib php5-xml
 
-RUN apk add curl && \
+ENV NODE_VERSION=v5.12.0 \
+    NPM_VERSION=3
+
+# Based on https://github.com/mhart/alpine-node
+RUN apk add --no-cache curl make gcc g++ linux-headers binutils-gold gnupg libstdc++ && \
+    gpg --keyserver ha.pool.sks-keyservers.net --recv-keys \
+      94AE36675C464D64BAFA68DD7434390BDBE9B9C5 \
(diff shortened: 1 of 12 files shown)


Dockerfile
```diff
@@ -26,6 +66,9 @@ RUN adduser -u 9000 -D -h /usr/src/app -s /bin/false app
 COPY . /usr/src/app
 RUN chown -R app:app /usr/src/app
 
+# ENV PURE_HASH=1
+# ENV CODECLIMATE_PROFILE=1
```
pbrisbin: Please remove before shipping.

lib/ccflay.rb
```diff
@@ -84,3 +84,25 @@ def pure_ruby_hash # :nodoc: see above
     hash
   end
 end
+
+class Sexp
```
pbrisbin: Is there a reason you re-open again, instead of including `#flatter` in the class definition opened at L38?

lib/ccflay.rb
```diff
@@ -84,3 +84,25 @@ def pure_ruby_hash # :nodoc: see above
...
+  def flatter
+    r = dup.clear
+
+    each do |s|
+      if s.is_a? Sexp
```
pbrisbin: :lipstick: Please use parens when invoking methods, as is our house style.

This applies throughout this block, but doesn't include `#<<`.

lib/ccflay.rb
```diff
@@ -84,3 +84,25 @@ def pure_ruby_hash # :nodoc: see above
...
+          r.concat ss
+        else
+          r << ss
+        end
+      else
+        r << s
+      end
+    end
+
+    r
```
pbrisbin: :lipstick: we prefer `#each_with_object` for cases like this

```rb
each_with_object(dup.clear) do |s, r|
  # r << ...
end
```

Dockerfile
zenspider: Sure... I'd like to leave the PURE_HASH as a TODO line... it does come out ahead in my measurements.

lib/ccflay.rb
zenspider: Yes, 1) it's my style. 2) the previous block needs to be pushed back to flay itself as it is a measurable win for everyone. It's even faster than ruby's default hash.

lib/ccflay.rb
zenspider: hey... if rubocop didn't ding it... 👅

lib/ccflay.rb
zenspider: I disagree with this for many reasons... but I think it doesn't matter because I'm gonna have to cache the old mass with `r.mass = mass` and that just doesn't fit with `each_with_object`.

Also... 15-18% slower?!? Yielding an extra parameter shouldn't cost THAT much.

```
Comparison:
            flatter1:        6.2 i/s
            flatter2:        5.3 i/s - 1.16x  slower
```

zenspider: I'm having a hard time believing it costs that much... I'd like to understand that better, but maybe not on your dime.

PR #167, opened 2017-03-21
Fix non-ruby speed problems (#166)

GordonDiggs:
Promoting @zenspider's changes to stable after some testing (it looks good!)


zenspider:
Nooooo!

GordonDiggs:
@zenspider what's up?

PR #168, opened 2017-03-21
Change the default remediation time to 30min

ABaldwinHunter:
In our system, 50K remediation points translated to 5 minutes of
developer work:
https://github.com/codeclimate/spec/blob/master/SPEC.md#remediation-points

We have decided that 30 minutes is a more accurate default base amount of
time needed to fix a duplication issue.

### DIFF
--- a/lib/cc/engine/analyzers/analyzer_base.rb
+++ b/lib/cc/engine/analyzers/analyzer_base.rb
@@ -13,7 +13,7 @@ module CC
           ::RuntimeError,
         ].freeze
 
-        BASE_POINTS = 1_500_000
+        BASE_POINTS = 300_000 # 30 minutes (5min = 50K)
 
         def initialize(engine_config:)
           @engine_config = engine_config
--- a/lib/cc/engine/analyzers/analyzer_base.rb
+++ b/lib/cc/engine/analyzers/analyzer_base.rb
@@ -13,7 +13,7 @@ module CC
           ::RuntimeError,
         ].freeze
 
-        POINTS_PER_MINUTE = 10_000 # Remediation points represent engineering time to resolve issue
+        POINTS_PER_MINUTE = 10_000 # Points represent engineering time to resolve issue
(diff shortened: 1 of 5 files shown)


lib/cc/engine/analyzers/analyzer_base.rb
```diff
@@ -13,7 +13,7 @@ class Base
           ::RuntimeError,
         ].freeze
 
-        BASE_POINTS = 1_500_000
+        BASE_POINTS = 300_000 # 30 minutes (5min = 50K)
```
pbrisbin: Since you decided to include some commentary, WDYT of making it executable, via something like: `POINTS_PER_MINUTE = 10_000; BASE_POINTS = 30 * POINTS_PER_MINUTE`?

For now we'd end up with the same 10K constant in a few code-bases but maybe they get consolidated soon.

ABaldwinHunter: Okay

ABaldwinHunter: @pbrisbin I also added `HOUR_POINTS` because I think hours are easier to think about for the size of remediation here - esp when we add the severity threshold of two hours. wdyt?

lib/cc/engine/analyzers/analyzer_base.rb
```diff
@@ -13,7 +13,9 @@ class Base
           ::RuntimeError,
         ].freeze
 
-        BASE_POINTS = 1_500_000
+        POINTS_PER_MINUTE = 10_000 # Points represent engineering time to remediate
+        HOUR_POINTS = 60 * POINTS_PER_MINUTE
+        BASE_POINTS = HOUR_POINTS / 2
```
pbrisbin: FWIW, this is a bit too fancy for my tastes. `30 * POINTS_PER_MINUTE` is perfectly clear IMO. Also `HOUR_POINTS` is grammatically odd and inconsistent (s/b `POINTS_PER_HOUR`).

PR #170, opened 2017-03-22
Introduce issue Severities

ABaldwinHunter:
In the context of technical debt, severity is in part a function of the amount
of engineering effort required to repair a  code issue. Here we
introduce "major" and "minor" severity levels mapping to amounts of
effort require to remediate an outbreak of duplicated code.

Duplicated code that takes more than 2 hours to repair (across all of its
occurrences), will have a severity of "major", and others will be assigned
a severity of "minor".

See the Code Climate SPEC for more information about severity and the permitted levels:
https://github.com/codeclimate/spec/blob/master/SPEC.md#issue://github.com/codeclimate/spec/blob/master/SPEC.md#issues

### DIFF
--- a/lib/cc/engine/analyzers/analyzer_base.rb
+++ b/lib/cc/engine/analyzers/analyzer_base.rb
@@ -21,7 +21,7 @@ module CC
           MINOR = "minor".freeze,
         ].freeze
 
-        SEVERITY_THRESHOLD = 120 * POINTS_PER_MINUTE # > 2 hours of work is major
+        MAJOR_SEVERITY_THRESHOLD = 120 * POINTS_PER_MINUTE
 
         def initialize(engine_config:)
           @engine_config = engine_config
@@ -60,7 +60,7 @@ module CC
         end
 
         def calculate_severity(points)
-          if points > SEVERITY_THRESHOLD
+          if points > MAJOR_SEVERITY_THRESHOLD
             MAJOR
           else
             MINOR
--- a/lib/cc/engine/analyzers/analyzer_base.rb
+++ b/lib/cc/engine/analyzers/analyzer_base.rb
@@ -16,6 +16,13 @@ module CC
         POINTS_PER_MINUTE = 10_000 # Points represent engineering time to resolve issue
         BASE_POINTS = 30 * POINTS_PER_MINUTE
 
+        SEVERITIES = [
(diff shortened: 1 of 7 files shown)


lib/cc/engine/analyzers/analyzer_base.rb
```diff
@@ -16,6 +16,13 @@ class Base
         POINTS_PER_MINUTE = 10_000 # Points represent engineering time to resolve issue
         BASE_POINTS = 30 * POINTS_PER_MINUTE
 
+        SEVERITIES = [
+          MAJOR = "major".freeze,
+          MINOR = "minor".freeze,
+        ].freeze
+
+        SEVERITY_THRESHOLD = 120 * POINTS_PER_MINUTE # > 2 hours of work is major
```
pbrisbin: FWIW: this variable should probably be called `MAJOR_SEVERITY_THRESHOLD`. Also, as you might expect, I find the comment unnecessary ;)

ABaldwinHunter: heh okay sounds fine on both fronts.

WDYT about the `MAJOR` and `MINOR` constants? wasn't sure if they needed prefixes?

pbrisbin: They're fine with me. You should know by now I'd've commented if not :)

brynary:
Just a quick Q RE: the title of this... Is "Violation" really an "Issue"? In our domain I think we use the term Issue, and I'm not sure I see a meaningful distinction where we want another noun here. Thoughts?

pbrisbin:
The new code here doesn't use the world violation, though the existing engine (apparently) does. We had a convo about naming in Slack earlier and everyone agrees Violation is wrong and Issue is right. We might still come across some old code using the wrong word (like this engine), but we're all good on direction AFAICT, and are addressing the existing misnaming when appropriate.

spec/cc/engine/analyzers/python/main_spec.rb
```diff
@@ -139,7 +142,7 @@ def engine_conf
       "config" => {
         "languages" => {
           "python" => {
-            "mass_threshold" => 4,
+            "mass_threshold" => 5,
```
ABaldwinHunter: this changed to get more interesting severity results - minor and major

PR #172, opened 2017-03-23
[WIP] Fix stack overflow

maxjacobson:
(need to confirm with QA and a test, but doesn't this look right?)

Example error:

```
/usr/src/app/lib/ccflay.rb:116:in `block in mass': stack level too deep (SystemStackError)
	from /usr/src/app/lib/ccflay.rb:116:in `each'
	from /usr/src/app/lib/ccflay.rb:116:in `inject'
	from /usr/src/app/lib/ccflay.rb:116:in `mass'
	from /usr/src/app/lib/ccflay.rb:118:in `block in mass'
	from /usr/src/app/lib/ccflay.rb:116:in `each'
	from /usr/src/app/lib/ccflay.rb:116:in `inject'
	from /usr/src/app/lib/ccflay.rb:116:in `mass'
	from /usr/src/app/lib/ccflay.rb:118:in `block in mass'
	 ... 1344 levels...
	from /usr/src/app/lib/cc/engine/analyzers/php/main.rb:21:in `transform_sexp'
	from /usr/src/app/lib/cc/engine/analyzers/reporter.rb:69:in `process_sexp'
	from /usr/src/app/lib/cc/engine/analyzers/reporter.rb:41:in `block in process_files'
	from /usr/src/app/lib/cc/engine/analyzers/file_thread_pool.rb:22:in `block (2 levels) in run
```

### DIFF
--- a/lib/ccflay.rb
+++ b/lib/ccflay.rb
@@ -115,7 +115,7 @@ class Sexp # TODO: push this back to flay
   def mass
     @mass ||= inject(1) { |t, s|
       if Sexp === s then
-        t + s.mass
+        t + s.old_mass
       else
         t
       end


pbrisbin:
I don't know how using `#old_mass` changes the semantics, or if that's actually the bug.

By my understanding of this code, this is blowing the stack because `self.each.to_a` (i.e. the elements over which `inject` is operating, I assume) must contain `self`. This code clearly doesn't expect that, I probably wouldn't expect it myself, but I'm not sure if that fact _should_ be expected (so we need another guard on the recursion, or a change like using `#old_mass`) or if there's a different bug elsewhere that's causing the list of elements to self-reference unexpectedly.

maxjacobson:
Yeeup. I may re-open if I understand this better, but closing for now

PR #180, opened 2017-04-13
Fix line numbers for duplicative hashes

maxjacobson:
Via an escalation

This seems like a bug in ruby_parser. This fix feels a little hacky, but it's a first draft.

cc @zenspider WDYT?


zenspider:
I just fixed this bug in ruby_parser. This should be reverted once I push a release.

maxjacobson:
@zenspider ah, awesome! Will you let us know when you've done that?

zenspider:
Working on it now

PR #182, opened 2017-04-13
Distinguish between identical and similar for PHP

maxjacobson:
We were always emitting issues as identical, even when they're only
similar.

In fact, our test covering the finding of identical issues featured some
code that wasn't actually identical, it was only similar.

The problem is that we were discarding information like the name of
variables and the values of string literals at the point when we transform
from a Node into a Sexp.

This change introduces the concept of an "identifier" for a node. If a node
has an identifier, and doesn't have any subnodes, that identifier should be
included in the Sexp. That way flay can tell whether your duplication
issues are actually identical or similar.

### DIFF
--- a/lib/cc/engine/analyzers/php/visitor.rb
+++ b/lib/cc/engine/analyzers/php/visitor.rb
@@ -183,22 +183,18 @@ module CC
           ALL_NODES.each do |type|
             eval <<-RUBY
               def visit_#{type}Node(node)
-                node.sub_nodes.map do |sub_node|
+                name = :#{type.gsub(/([a-z])([A-Z])/, '\1_\2').downcase}
+                sub = node.sub_nodes.map do |sub_node|
                   sub_node.accept(self)
                 end
-              end
-            RUBY
-          end
-        end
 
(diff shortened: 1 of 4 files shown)


lib/cc/engine/analyzers/php/visitor.rb
```diff
@@ -182,21 +182,17 @@ def accept(target)
 
           ALL_NODES.each do |type|
             eval <<-RUBY
-              def visit_#{type}Node(node)
-                node.sub_nodes.map do |sub_node|
+              def visit_#{type}Node(o)
```
chrishulton: can you use something more descriptive than `o`?  (i know that's what was there before)

maxjacobson:
@chrishulton After I opened this PR last night, @zenspider asked me why not just include all of the node's attributes in the sexp. I'm going to try to do that, and if I change this PR meaningfully I might ask for more review

maxjacobson:
I'm going to punt on that for now. I'm not totally clear on how to structure the Sexp to include arbitrary metadata, and I think this is a meaningful enough improvement that it's worth shipping as is.

PR #183, opened 2017-04-14
Bundle after node build

wfleming:
The intention of grouping everything in one layer is to optimize final storage
sizes for large deps we remove, like compiler toolchains.

We don't currently depend on any Ruby extensions in the gems we're using, so it
should be fine to `bundle install` in a later layer. This can speed up iteration
& build times when changing gem deps.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -48,13 +48,14 @@ RUN apk add --no-cache curl make gcc g++ linux-headers binutils-gold gnupg libst
       find /usr/lib/node_modules/npm -name test -o -name .bin -type d | xargs rm -rf; \
     fi && \
     gem install bundler --no-ri --no-rdoc && \
-    (cd /usr/src/app && bundle install -j 4) && \
     (cd /usr/src/app && curl -sS https://getcomposer.org/installer | php) && \
     apk del --purge curl make gcc g++ linux-headers binutils-gold gnupg ${DEL_PKGS} && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -48,14 +48,15 @@ RUN apk add --no-cache curl make gcc g++ linux-headers binutils-gold gnupg libst
...
     gem install bundler --no-ri --no-rdoc && \
-    (cd /usr/src/app && bundle install -j 4) && \
     (cd /usr/src/app && curl -sS https://getcomposer.org/installer | php) && \
     apk del --purge curl make gcc g++ linux-headers binutils-gold gnupg ${DEL_PKGS} && \
     rm -rf ${RM_DIRS} /node-${NODE_VERSION}* /usr/share/man /tmp/* /var/cache/apk/* \
       /root/.npm /root/.node-gyp /root/.gnupg /usr/lib/node_modules/npm/man \
       /usr/lib/node_modules/npm/doc /usr/lib/node_modules/npm/html \
       /usr/lib/node_modules/npm/scripts
 
+RUN cd /usr/src/app && bundle install -j 4
```
GordonDiggs: Why not put this after the `WORKDIR` line?

zenspider:
json is a C extension.

And this change prevents my other PR from working... you can't use ANY tooling to debug/profile/etc that requires a C extension.

A better move would be to separate the whole thing and use `docker build --squash`.

PR #184, opened 2017-05-02
Quality-Model Duplication Tweaks

chrishulton:
* Updates issue description to `[Similar/Identical] blocks of code found in N locations. Consider refactoring.` which is in-line with the ruby-complexity messages
* Changes default thresholds for Ruby:
  * Mass threshold: Bumped from `18` to `25` (this is what was set in app's codeclimate.yml)
  * Base remediation points: Lowered from `300_000` (30 min) to `150_000` (15 min)
  * Points per overage: Lowered from `100_000` (10 min) to `20_000` (2 min)

Before this change, the total remediation time for duplication issues in app was 310 hours, or ~39 business days (!).  After, it's 41 hours, or ~5 business days, which is more reasonable. Will likely have another pass to tweak after this is merged.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ image:
 	docker build --rm -t $(IMAGE_NAME) .
 
 test: image
-	docker run --tty --interactive --rm $(IMAGE_NAME) bundle exec rake
+	docker run --tty --interactive --rm $(IMAGE_NAME) bundle exec rspec $(RSPEC_ARGS)
 
 citest:
 	docker run --rm $(IMAGE_NAME) bundle exec rake
--- a/spec/cc/engine/analyzers/javascript/main_spec.rb
+++ b/spec/cc/engine/analyzers/javascript/main_spec.rb
@@ -21,7 +21,7 @@ RSpec.describe CC::Engine::Analyzers::Javascript::Main, in_tmpdir: true do
 
       expect(json["type"]).to eq("issue")
       expect(json["check_name"]).to eq("Identical code")
-      expect(json["description"]).to eq("Avoid identical blocks of code (3 locations). Consider refactoring.")
+      expect(json["description"]).to eq("Identical blocks of code found in 3 locations. Consider refactoring.")
       expect(json["categories"]).to eq(["Duplication"])
       expect(json["location"]).to eq({
         "path" => "foo.js",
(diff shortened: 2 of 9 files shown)


spec/cc/engine/analyzers/ruby/main_spec.rb
```diff
@@ -29,15 +29,15 @@ def self.sub_degrees(str)
               before { subject.type = 'ruby' }
 
               it 'returns true' do
-                expect(subject.ruby?).to be true
+                10.times { |i| if i < 5; if i % 2 == 0; subject.increase_mass!; end; end }; expect(subject.ruby?).to be true
```
chrishulton: i put this all on one line to not have to change all the locations in the specs. it's just to bump the complexity over the new threshold of 25.

Makefile
```diff
@@ -6,7 +6,7 @@ image:
 	docker build --rm -t $(IMAGE_NAME) .
 
 test: image
-	docker run --tty --interactive --rm $(IMAGE_NAME) bundle exec rake
+	docker run --tty --interactive --rm $(IMAGE_NAME) bundle exec rspec $(RSPEC_ARGS)
```
ABaldwinHunter: nice

spec/cc/engine/analyzers/javascript/main_spec.rb
```diff
@@ -21,7 +21,7 @@
 
       expect(json["type"]).to eq("issue")
       expect(json["check_name"]).to eq("Identical code")
-      expect(json["description"]).to eq("Identical code found in 2 other locations (mass = 11)")
+      expect(json["description"]).to eq("Identical blocks of code found in 3 locations. Consider refactoring.")
```
ABaldwinHunter: I like the change to mention total blocks of code, though it seems like a product question. 

Are we going to expose the mass elsewhere?

ABaldwinHunter:
> Before this change, the total remediation time for duplication issues in app was 310 hours, or ~39 business days (!). After, it's 41 hours, or ~5 business days, which is more reasonable. 

Excellent description! :)

I personally like having the mass in the UI but I'm guessing that came from a Product decision earlier, and these changes are just for us for now.

So LGTM!

Makefile
ABaldwinHunter: we could probably make this in master as well. total side note: was thinking we might want to merge master into this branch periodically

spec/cc/engine/analyzers/javascript/main_spec.rb
chrishulton: Yeah, this was discussed with Bryan. Now that we show the "time to fix" underneath the issue, it's less important to show the mass (higher remediation time => bigger mass). 

The mass is also still included in the "content" field.

Makefile
chrishulton: yeah, pat and i were talking about this briefly earlier.  i re-based master on this branch today, so it's at least up to date now.

PR #185, opened 2017-05-03
Adds fix for begin and end lines

chrishulton:
A fix was added to better compute the line min in this commit: https://github.com/codeclimate/codeclimate-duplication/commit/0884d119a7776b3946349410e7229ff4014b6fd1

However there are still buggy cases such as when the lines are equal.  The idea here is to always take the minimum of the patched `sexp.line_min` and the previous `sexp.line`, and the max of the patched `sexp.line_max` and the previous `sexp.end_line` and use that for the location.

This is an example of the location not working before the change:
image

And after:
image

It also fixes a line number spec with a `TODO` about a bug in the parser, now returning the values before that comment was added, which seems promising.
https://github.com/codeclimate/codeclimate-duplication/commit/fd6ff180b5ef030f2d0dcc9a74b162990fdf7abe#diff-e5c5455c0ff7cf41dbcfc36816e07075

### DIFF
--- a/lib/cc/engine/analyzers/sexp_lines.rb
+++ b/lib/cc/engine/analyzers/sexp_lines.rb
@@ -16,12 +16,8 @@ module CC
         attr_reader :root_sexp
 
         def calculate
-          @begin_line = root_sexp.line
-          @end_line = root_sexp.end_line || root_sexp.line_max
-
-          if @end_line < @begin_line
-            @begin_line = root_sexp.line_min
-          end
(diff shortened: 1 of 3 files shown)


maxjacobson:
I'd be curious to benchmark this change. `line_min` and `line_max`, I think, are expensive and to be avoided if possible. If it's not a big difference, cool

chrishulton:
@maxjacobson I ran a few tests on app (old vs this) and did not see a noticeable change in time.  I think that makes sense since this is only run for violations.

maxjacobson:
Good point

PR #189, opened 2017-06-05
Respect QM checks

ABaldwinHunter:
@wfleming currently working on tests. opening early for feedback on the approach.

### DIFF
--- a/lib/cc/engine/analyzers/engine_config.rb
+++ b/lib/cc/engine/analyzers/engine_config.rb
@@ -5,6 +5,9 @@ module CC
     module Analyzers
       class EngineConfig
         DEFAULT_COUNT_THRESHOLD = 2
+        IDENTICAL_CODE_CHECK = "identical-code".freeze
+        SIMILAR_CODE_CHECK = "similar-code".freeze
+
         InvalidConfigError = Class.new(StandardError)
 
         def initialize(hash)
@@ -25,6 +28,14 @@ module CC
 
         def concurrency
           config.fetch("config", {}).fetch("concurrency", 2).to_i
+        end
+
+        def identical_code_check_enabled?
+          enabled?(IDENTICAL_CODE_CHECK)
+        end
+
+        def similar_code_check_enabled?
+          enabled?(SIMILAR_CODE_CHECK)
         end
 
         def mass_threshold_for(language)
@@ -95,6 +106,14 @@ module CC
           else
             raise InvalidConfigError, "#{entry.inspect} is not a valid language entry"
           end
+        end
+
+        def checks
+          config.fetch("config", {}).fetch("checks", {})
+        end
+
+        def enabled?(check)
+          checks.fetch(check, {}).fetch("enabled", true)
         end
       end
     end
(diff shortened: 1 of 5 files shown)


lib/cc/engine/analyzers/engine_config.rb
```diff
@@ -5,6 +5,9 @@ module Engine
     module Analyzers
       class EngineConfig
         DEFAULT_COUNT_THRESHOLD = 2
+        IDENTICAL_CODE_CHECK = "identical_code".freeze
```
wfleming: 💄 these are dashes in the config, not underscores. https://github.com/codeclimate/rfc/pull/79

lib/cc/engine/analyzers/engine_config.rb
```diff
@@ -20,13 +23,29 @@ def include_paths
...
         def concurrency
           config.fetch("config", {}).fetch("concurrency", 2).to_i
         end
 
+        def checks
+          config.fetch("config", {}).fetch("checks", {})
+        end
+
+        def enabled?(check)
+          checks.fetch(check, {}).fetch("enabled", "false").to_s.casecmp("true").zero?
```
wfleming: 1. The values in the config will be real booleans, not strings. (Bools in the YAML `config` subkey are strings today because of weirdness in `cc-yaml`, but that has no bearing on the new stuff.)

2. Shouldn't checks default to enabled, not disabled?

lib/cc/engine/analyzers/engine_config.rb
ABaldwinHunter: thanks! good catch

lib/cc/engine/analyzers/engine_config.rb
ABaldwinHunter: 1. GTK - will update!
2. I could imagine arguments for either behavior but defaulting to enabled sounds fine.
(makes the specs easier as well :) )

spec/cc/engine/analyzers/javascript/main_spec.rb
```diff
@@ -124,6 +124,14 @@
   def engine_conf
     CC::Engine::Analyzers::EngineConfig.new({
       'config' => {
+        'checks' => {
+          'similar-code' => {
+            'enabled' => true,
+          },
+          'identical-code' => {
+            'enabled' => true,
```
ABaldwinHunter: this change isn't needed for the specs to pass, since the checks are turned on by default, but there may be some value in having the example.

lib/cc/engine/analyzers/engine_config.rb
```diff
@@ -20,13 +23,21 @@ def include_paths
         end
 
         def languages
-          config.fetch("languages", {})
+          config.fetch("languages", ["ruby"])
```
wfleming: This doesn't seem necessary here? The QM branch of CLI will provide this as the default anyway: https://github.com/codeclimate/codeclimate/blob/quality-model/lib/cc/config/json.rb#L40. It doesn't seem necessary for the check-config changes here, so I think we can keep this as-is.

spec/cc/engine/analyzers/javascript/main_spec.rb
wfleming: I hope we're going to exercise both the default (no `config.checks` key) as well as a config that overrides that default?

Issue #89, opened 2016-01-28
React's propTypes often marked as duplicates


jdelStrother:
Hi - a common source of false-positives in our CC duplication results comes from React's propType structures.  For example:

```js
  propTypes: {
    playBtnState: React.PropTypes.string.isRequired,
    height: React.PropTypes.number.isRequired,
    drawBack: React.PropTypes.bool
  },
```

is a 'duplicate' of:

```js
  propTypes: {
    clipStore: React.PropTypes.object.isRequired,
    firstClipId: React.PropTypes.number.isRequired,
    includeDetails: React.PropTypes.bool
  },
```

Can the engine be improved to avoid this, or is there anything we can do in our codebase that would help?   Excluding each fingerprint on a case-by-case basis is getting tiresome.

ABaldwinHunter:
@jdelStrother Thank you for the feedback! We are working to improve our Duplication engine's analysis of JavaScript and make results more savvy.

I'm going to tag this issue as `Enhancement` and add its consideration to our roadmap.  

In the meantime, one workaround might be to to experiment with upping the `mass` threshold for JavaScript duplication in your `.codeclimate.yml` config settings:

```
engines:
  duplication:
    enabled: true
    config:
      languages:
         javascript:
           mass_threshold: 50
```

Our default for JavaScript is 40. Increasing the threshold should make the detection of duplication less sensitive.

netes:
Very similar situation occurs when checking Ember imports:

```import Ember from 'ember';
import OnboardingButton from 'app/mixins/onboarding-button';

const { Component } = Ember;```

is a duplication of something like this ie.:
```import Ember from 'ember';
import UnauthenticatedRouteMixin from 'ember-simple-auth/mixins/unauthenticated-route-mixin';

const { Route } = Ember;

export default Route.extend(UnauthenticatedRouteMixin);```

This affects overall score a lot :(

wfleming:
@netes @jdelStrother Sorry for the frustration here. Right now [the duplication algorithm](https://docs.codeclimate.com/docs/duplication-concept) effectively compares syntax nodes based on both what *kind* of expression they represent, as well as the *contents* of those expressions. When the kinds of the expressions are the same, but the contents vary (which looks to be the case here), the engine considers that "similar" duplication to distinguish it from "identical" duplication.

[...]

PR #190, opened 2017-06-14
Add language-specific node filtering

zenspider:
99% of the work wound up in sexp_processor, which got released to beta yesterday. This has minor patches against Flay that will get pushed upstream and released.

This also has changes to parsing strategy. In my measurements, if a (especially foreign language) file can't parse in 10 seconds, it can't parse in 5 minutes (the default timeout for CommandLineRunner), basically chewing up 1/3rd of the overall default timeout for no benefit. By changing them to 10s, I timeout on only the most egregious files but now finish w/o a timeout error on all the projects.

The only thing I find really icky is the `new_nodes` array in lib/ccflay.rb. This is needed for the stable hash algorithm. It is probably missing node types and it would be nice if there was a comprehensive list, but the foreign parsers are messy and relatively undocumented as such. (I'd highly recommend adding a rewrite phase to cleane and speed things up even further).


zenspider:
As extra datapoints:

1. the big_files project used to time out at 900s, even with only source 3 files in it.
2. Profiling and changes to both `codeclimate` and `codeclimate-duplication` dropped runtime down to ~300s.
3. Filtering (everything) drops the time down to 30s.

spec/cc/engine/analyzers/ruby/main_spec.rb
```diff
@@ -197,5 +235,18 @@ def self.from_remediation_amount(amount)
     def engine_conf
       EngineConfig.new({})
     end
+
+    def filtered_engine_conf *patterns
+      EngineConfig.new({
```
dblandin: Is this indentation intentional?

spec/cc/engine/analyzers/sexp_lines_spec.rb
```diff
@@ -49,8 +49,8 @@ module CC::Engine::Analyzers
 
         expect(locations.count).to eq 2
 
-        expect([locations[0].begin_line, locations[0].end_line]).to eq([2, 7])
-        expect([locations[1].begin_line, locations[1].end_line]).to eq([11, 16])
+        expect([locations[0].begin_line, locations[0].end_line]).to eq([1, 7])
+        expect([locations[1].begin_line, locations[1].end_line]).to eq([10, 16])
```
dblandin: I'm curious what caused these line locations to change.

lib/cc/engine/analyzers/engine_config.rb
```diff
@@ -27,6 +27,12 @@ def concurrency
           config.fetch("config", {}).fetch("concurrency", 2).to_i
         end
 
+        def filters_for(language)
+          fetch_language(language).fetch("filters", []).map { |s|
+            Sexp::Matcher.parse s
+          }
```
dblandin: What does the `s` variable represent? Does it stand for string? What do you think of fleshing it out to "filter" if we're mapping over filters?

spec/cc/engine/analyzers/ruby/main_spec.rb
zenspider: Intentional, yes. Good? probably not. It's how emacs does it by default (more like lisp indentation).

spec/cc/engine/analyzers/sexp_lines_spec.rb
zenspider: Probably fixes to ruby_parser. I have a release coming out soon that makes more of these fixes so there might be more failed tests coming up.

lib/cc/engine/analyzers/engine_config.rb
zenspider: :+1:

zenspider:
@dblandin:

> Is the default parsing timeout value of 300 useful anymore? Would it make sense to change that default to 10 and override it to 30 within the ruby parser?

I suspect it probably never made sense. 30 or 10 are more more sensible defaults based on the numbers I've gathered. I'd be fine dropping the numbers... but it might be better if data were gathered so that statistics could decide what the numbers should be.

> Would it make sense to define some default filters when none are defined with the codeclimate yml configuration?

Absolutely. But I suspect at this time it would be premature to decide what those should be. @ABaldwinHunter and others would have more experienced/valid opinions on this matter than I would.

dblandin:
This is available on the beta channel now:

```yaml
engines:
  duplication:
    enabled: true
    channel: beta
    config:
      languages:
      - ruby
```

gettinToasty:
I just tried using this on the beta channel but I'm getting this error:
```
Invalid configuration.
Errors:
- syntax error: (<unknown>): mapping values are not allowed in this context at line 15 column 19
```
Is this not quite ready for use?

dblandin:
@gettinToasty Would you mind sharing your `.codeclimate.yml` file?

gettinToasty:
@dblandin I managed to fix that issue-- the new readme says the format should be
```
config:
    languages:
        javascript:
          filters:
```
but it should be:
```
config:
      languages:
      - javascript:
          filters:
```

i am having some trouble with the pattern matching though. my main problem is propTypes being marked as duplicates so i tried `- "(object (lit -) (PropTypes _) ___)"` as well as `- "(object ___)"` but didn't get any fixed results

dblandin:
> i am having some trouble with the pattern matching though. my main problem is propTypes being marked as duplicates so i tried - "(object (lit _) (PropTypes _) ___)" as well as - "(object ___)" but didn't get any fixed results

@zenspider should be able to help with the pattern syntax ^

@gettinToasty Could you provide an example case of detected duplication you're trying to filter out?

gettinToasty:
@dblandin sure thing! here are some lightly modified code snippets:
```
EventBlock.propTypes = {
  heading: PropTypes.string.isRequired,
  sub_heading: PropTypes.string.isRequired,
  description: PropTypes.string.isRequired,
  event_description: PropTypes.string.isRequired,
```
```
VideoList.propTypes = {
  heading: PropTypes.string.isRequired,
  list: PropTypes.string.isRequired,
  subtext: PropTypes.string.isRequired,
  closing_line: PropTypes.string.isRequired,
```

which have a Mass 90 duplication issue

zenspider:
Hrm... wrt the config, here is what I'm using:

```
engines:
  duplication:
    enabled: true
    config:
      debug: true
      languages:
        ruby:
          filters:
          - "([m /./] ___)"
        javascript:
          filters:
          - "([m /./] ___)"
        php:
          filters:
          - "([m /./] ___)"
```

(intentionally nuking EVERYTHING, to see full coverage and its corresponding performance differences)

zenspider:
OK. WRT the code... this is a doozy. I am not fond of the AST that comes back from javascript, but after our munging, this is how it looks:

```ruby
s(:Program,
 :module,
 s(:body,
  s(:ExpressionStatement,
   :expression,
   s(:AssignmentExpression,
    :"=",
    :left,
    s(:MemberExpression,
     :object,
     s(:Identifier, :EventBlock),
     :property,
     s(:Identifier, :propTypes)),
    s(:right,
     :ObjectExpression,
     s(:properties,
      s(:ObjectProperty,
       :key,
       s(:Identifier, :heading),
       :value,
       s(:MemberExpression,
        :object,
        s(:MemberExpression,
         :object,
         s(:Identifier, :PropTypes),
         :property,
         s(:Identifier, :string)),
        :property,
        s(:Identifier, :isRequired))),
...
```

so you probably want something more like:

```
(AssignmentExpression _ _ (MemberExpression object _ property _) ___)
```

OR:

```
(right ObjectExpression (properties ___))
```

gettinToasty:
@zenspider would something like `(AssignmentExpression _ ([m /propTypes$/]))` work as well?

zenspider:
I'd also really like some feedback on how you (or anyone) thinks that you should be able to see the AST to figure that out. Right now I have a patch on CCD that I run inside docker that dumps it. It is _incredibly verbose_ for javascript (as hinted, the AST is ginormous).

I could probably have this set up to output to the debug log if `CODECLIMATE_DEBUG` is set... but I'm not sure codeclimate employees want to see that normally. I'm totally open to suggestion.

gettinToasty:
@zenspider as someone completely new to this syntax there's no way i would have been able to get anywhere close without seeing the AST haha, but that first expression you gave seems to do the trick really well

zenspider:
Yeah... The JS parser in particular is a little crazy... I want to expose SOMETHING to make it easier, just not sure what it should be yet...

Maybe the filtering should just be against the structure of the tree and not the tree itself? @GordonDiggs ?

GordonDiggs:
@zenspider can you explain the different options there a bit?

zenspider:
@GordonDiggs I've been mulling over this for a week and I think it'd be a mistake... 

So, the difference... given some code:

```ruby
def x(y)
  y + 1
end
```

The raw parse tree looks like:

```ruby
s(:defn, :x, s(:args, :y), s(:call, s(:lvar, :y), :+, s(:lit, 1)))
```

and the structural tree looks like:

```ruby
s(:defn, s(:args), s(:call, s(:lvar), s(:lit)))
```

which is easier to read and perhaps easier to write filters against, but it also means you _can't_ filter on, for example, method names.

I can totally imagine people filtering on something like: `(defn [m /^test_/] ___)`

zenspider:
I found a tool: http://astexplorer.net It lets you visualize and explore the AST in a few languages, with the main focus being javascript. It might help (although right now we're massaging our AST for performance reasons which might make it a bit trickier).

zenspider:
Just pushed PR #193 that allows you to dump the AST to better help with writing filters.

dblandin:
I'm pretty sure that this has been merged upstream, so I'm going to go ahead and close this PR. Let me know if that's not the case!

gettinToasty:
@dblandin i believe this functionality is only on the beta branch and not the main branch; i put up a new PR trying to use filters first with the beta branch which had what looked like some engine changes as well (new issues discovered) but when i removed `channel: beta` it took away all of the duplication fixes as a result of the filter

hiattp:
@gettinToasty I know this ticket is old but we're trying to filter out `PropTypes` similarity warnings as well...happen to remember what filter you ended up using? #89 suggests that changing the `mass_threshold` may be the best bet, but @zenspider's filtering looks more robust if PropType definitions are the only issue

PR #197, opened 2017-07-20
Replace patrick with raw docker pull

GordonDiggs:
We've started to use this technique elsewhere

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -4,14 +4,7 @@ machine:
 
 dependencies:
   override:
-    - >
-      docker run
-      --env CIRCLE_BRANCH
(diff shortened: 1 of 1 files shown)


GordonDiggs:
this isn't working like i had hoped :/

pbrisbin:
I'd be happy to help with this if you want, I'd like to move all our stuff to the approach I did in (e.g.) https://github.com/codeclimate/finalizer/pull/227 where CI releases to whatever registries (including DH) and pre-pulls from DH as well. For engines, cli, etc we'd want to remove GCR<>DH relay Shipbot does.

PR #200, opened 2017-07-20
Use parser

nporteschaikin:
I think the best way to assess this PR is to observe the diff of each
language's analyzer spec. I needed to adjust the asserted remediation
points, masses, etc. to get the specs to pass with the new ASTs.


nporteschaikin:
cc/ @ABaldwinHunter

spec/cc/engine/analyzers/node_translator_spec.rb
```diff
@@ -0,0 +1,40 @@
+require "spec_helper"
+require "cc/engine/analyzers/node_translator"
+
+RSpec.describe(CC::Engine::Analyzers::NodeTranslator) do
```
pbrisbin: `NodeTranslator` is overloaded IMO, because `CC::Parser::NodeTranslator` exists. I liked this better when it was named `SexpBuilder`, or something that indicates that it's `Node -> Sexp` (because `NodeTranslator` implies `Node -> Node` to me). Can you describe a little bit what led you to this name?

nporteschaikin: Totally forgot about `CC::Parser::NodeTranslator`. My thinking was simple: the class translates a node. I don't necessarily think it implies `Node -> Node` but I see your point. Happy to change it.

pbrisbin: > My thinking was simple: the class translates a node

True, but not specific (IMO), not a huge deal. Maybe `NodeTranslator#to_sexp` would be a good compromise?

nporteschaikin: Already moved it back :)

ABaldwinHunter:
These changes look really good to me!

One tiny CX item to consider (and probably not be concerned with) is that since changes in mass impact fingerprints, there could be a minor impact on users who had tried to target issues.

i.e.
-  `exclude_fingerprints` in a .codeclimate.yml (_if_ we still care about that in the QM world?)
- Some users might no longer have certain issues ignored / confirmed / wontfix.

lib/cc/engine/analyzers/analyzer_base.rb
```diff
@@ -30,13 +22,9 @@ def initialize(engine_config:)
...
-          if (skip_reason = skip?(file))
-            $stderr.puts("Skipping file #{file} because #{skip_reason}")
-          else
-            process_file(file)
-          end
+          build_sexp(node(file), file)
         rescue => ex
-          if RESCUABLE_ERRORS.map { |klass| ex.instance_of?(klass) }.include?(true)
+          if ex.is_a?(CC::Parser::Client::HTTPError)
             $stderr.puts("Skipping file #{file} due to exception (#{ex.class}): #{ex.message}\n#{ex.backtrace.join("\n")}")
```
ABaldwinHunter: maybe we should update this to better match the error handling in structure? https://github.com/codeclimate/codeclimate-structure/commit/02b008264f04472ac38b50742405264ce41629b1

wfleming:
I definitely think we should do this as a beta we can dogfood on our own repos to work out any kinks. That would also separate the needed work of tuning constants from the review & stress testing of code changes.

nporteschaikin:
@wfleming Great idea.

nporteschaikin:
We've put this on `channel/quality-model-beta`; I think we should close this and work off that branch. WDYT, @pbrisbin?

pbrisbin:
Sure, WFM.

PR #207, opened 2017-08-28
Use codeclimate-parser backend

dblandin:
Setup a `java` channel to pull in our new parser backend. This keeps all of the core language implementation intact and unchanged while we build out java support.

This PR makes the parser available via the base image. A following PR will add the Java support implementation.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@ RUN apt-get update && \
   RUNLEVEL=1 apt-get install --yes --no-install-recommends \
     python2.7
 
-COPY Gemfile* package.json ./
+COPY Gemfile* package.json package-lock.json ./
 COPY vendor/php-parser/composer* ./vendor/php-parser/
 
 RUN bundle install --jobs 4 --quiet && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,73 +1,27 @@
-FROM ruby:2.3-alpine
+FROM codeclimate/codeclimate-parser:b264
+LABEL maintainer="Code Climate <hello@codeclimate.com>"
+
+# Reset from base image
+USER root
 
 WORKDIR /usr/src/app/
-
-COPY Gemfile /usr/src/app/
-COPY Gemfile.lock /usr/src/app/
-
-COPY vendor/php-parser/composer.json /usr/src/app/vendor/php-parser/
-COPY vendor/php-parser/composer.lock /usr/src/app/vendor/php-parser/
-
-COPY package.json /usr/src/app/
-
-RUN apk update && apk add python python3 php5-phar php5-openssl php5-cli php5-json php5-zlib php5-xml git
-
-ENV NODE_VERSION=v5.12.0 \
-    NPM_VERSION=3 \
-    CONFIG_FLAGS="--fully-static" \
-    DEL_PKGS="libstdc++" \
-    RM_DIRS=/usr/include
-
-WORKDIR /
-
-# sed line below is for aufs: https://github.com/npm/npm/issues/13306#issuecomment-236876133
-
-# https://github.com/npm/npm/pull/10373#issuecomment-195742307
-
-# Based on https://github.com/mhart/alpine-node
-RUN apk add --no-cache curl make gcc g++ linux-headers binutils-gold gnupg libstdc++ && \
-    gpg --keyserver ha.pool.sks-keyservers.net --recv-keys \
-      94AE36675C464D64BAFA68DD7434390BDBE9B9C5 \
-      FD3A5288F042B6850C66B31F09FE44734EB7990E \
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,74 +1,43 @@
...
+    php php-common php-cli php-fpm php-xml composer \
+    python python2.7 python-pip python-setuptools \
+    ruby2.3 ruby2.3-dev bundler
+
+COPY Gemfile* package.json /usr/src/app/
+COPY vendor/php-parser/composer* /usr/src/app/vendor/php-parser/
+
+RUN bundle install --jobs 4 && \
+    cd /usr/src/app/vendor/php-parser && composer install --no-interaction && \
+    npm install
```
maxjacobson: is this `npm install`-ing inside of the php directory?

package-lock.json
```diff
@@ -0,0 +1,13 @@
+{
```
maxjacobson: What's up with this file? Do we want it? If we do, I think we should `COPY` it in the dockerfile and document how to update it over time

Dockerfile
```diff
@@ -1,74 +1,43 @@
...
+  curl --location --silent https://deb.nodesource.com/setup_8.x | bash - && \
+  mkdir -p /usr/share/man/man1 && \
+  RUNLEVEL=1 apt-get install --yes --no-install-recommends \
+    dumb-init \
+    nginx \
+    nodejs \
+    openjdk-8-jdk-headless maven \
+    php php-common php-cli php-fpm php-xml composer \
+    python python2.7 python-pip python-setuptools \
+    ruby2.3 ruby2.3-dev bundler
```
maxjacobson: Are some of these things maybe already available via the base image and don't need to be re-installed? I suppose it doesn't hurt to re-install them

package-lock.json
dblandin: I'm including it for now because `npm install` outputs a warning when it writes the file. We'll be removing this particular `npm install` when we move over to the parser soon, so I'm inclined to punt on the documentation. Added that `COPY` argument though :+1:

Dockerfile
dblandin: Ahhh, yes, most of these are redundant. Cleaned up!

Dockerfile
```diff
@@ -1,73 +1,27 @@
...
 
-WORKDIR /usr/src/app/
+COPY Gemfile* package.json package-lock.json ./
+COPY vendor/php-parser/composer* /usr/src/app/vendor/php-parser/
 
-RUN mv composer.phar /usr/local/bin/composer
-RUN cd /usr/src/app/vendor/php-parser/ && composer install --prefer-source --no-interaction
-RUN npm install
+RUN bundle install --jobs 4 --quiet && \
+    composer install --no-interaction --quiet --working-dir /usr/src/app/vendor/php-parser && \
```
maxjacobson: `working-dir` nice

maxjacobson: these `--quiet`s are nice too, they definitely make the output less noisy

Dockerfile
dblandin: Ahh, good point. I don't think so, but I'm now taking advantage of a `composer install` flag that sets the working directory, just for the composer dependency installation.

Dockerfile
```diff
@@ -1,73 +1,27 @@
-FROM ruby:2.3-alpine
+FROM codeclimate/codeclimate-parser:java-beta
```
maxjacobson: if our channel release strategy is to make the tag based on the branch name, and not the build number, how do we bump the base image?

This looks correct based on our parser release strategy, but I'm just curious how we're planning to handle that...

Dockerfile
```diff
@@ -1,73 +1,27 @@
...
 
-WORKDIR /usr/src/app/
+COPY Gemfile* package.json package-lock.json ./
+COPY vendor/php-parser/composer* ./vendor/php-parser/
 
-RUN mv composer.phar /usr/local/bin/composer
-RUN cd /usr/src/app/vendor/php-parser/ && composer install --prefer-source --no-interaction
-RUN npm install
+RUN bundle install --jobs 4 --quiet && \
+    composer install --no-interaction --quiet --working-dir ./vendor/php-parser && \
```
maxjacobson: I like these `--quiet`s!

Dockerfile
dblandin: That's a good point. We probably shouldn't be using a `java-beta` tag for the codeclimate-parser image. I'll make some changes there, so we can use a `b###` tag here as well.

dblandin: Updated the `java` branch of `codeclimate-parser` to push `b###`-tagged images that we can use as the base for this channel: https://circleci.com/gh/codeclimate/codeclimate-parser/264

Dockerfile
dblandin: ssshhh

PR #208, opened 2017-08-28
Add initial Java duplication support

dblandin:
Adds a new Java language strategy, utilizing the new parser backend.

Kudos and credit to @ABaldwinHunter and @nporteschaikin on the `SexpBuilder` implementation!! 👍


lib/cc/engine/analyzers/java/main.rb
```diff
@@ -0,0 +1,35 @@
...
+
+module CC
+  module Engine
+    module Analyzers
+      module Java
+        class Main < CC::Engine::Analyzers::Base
+          LANGUAGE = "java".freeze
+          PATTERNS = ["**/*.java"].freeze
+          DEFAULT_MASS_THRESHOLD = 40
+          POINTS_PER_OVERAGE = 10_000
```
ABaldwinHunter: will we tune these eventually or feel happy? curious about how we chose these starting out?

dblandin: I'm sure we'll tune these but I think it'll be easier to flesh out the examples in the QM engines-test repo, and test on .com with forked Java repos then trying to figure out the right values locally during this initial PR.

lib/cc/engine/analyzers/java/main.rb
```diff
@@ -0,0 +1,35 @@
...
+          REQUEST_PATH = "/java".freeze
+          TIMEOUT = 300
+
+          private
+
+          def process_file(file)
+            node = ProcessedSource.new(file, REQUEST_PATH).ast
+
+            SexpBuilder.new(node, file).build
+          rescue Timeout::Error
```
ABaldwinHunter: is it likely we'll have this issue when using the new parser? I might expect a non 200 status code instead like, 502 or 503?

dblandin: Ahh, good point. It looks like we only use a Timeout block within the command line runner, which we aren't using for this strategy. I'll remove this rescue 👍

Gemfile
```diff
@@ -5,6 +5,9 @@ gem "concurrent-ruby", "~> 1.0.0"
 gem "flay"
 gem "json"
 
+gem "codeclimate-parser-client",
+  path: "/home/app/codeclimate-parser-client"
```
ABaldwinHunter: this is interesting. were we unable to use the parser-client without requiring here?

dblandin: The require statement you might be looking for is `require "cc/parser"` which is in this diff.

PR #209, opened 2017-08-29
Add QA script

maxjacobson:
The main difference from the other one is that this engine does require a configuration file.

Haven't enabled this for CI yet (there's a separate issue tracking that)

### DIFF
--- a/qa/bin/clone_and_test_examples
+++ b/qa/bin/clone_and_test_examples
@@ -11,6 +11,7 @@ if [ -n "$QA" ] || test_on_branch $CIRCLE_BRANCH; then
 
   examples="$(pwd)/$1"
   image_name="$2"
+  config_template_path="$(pwd)/$1/../config.template.json"
   config_path="$examples/config.json"
   mkdir -p "$examples"
   cd "$examples"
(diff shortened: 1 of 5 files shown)


maxjacobson:
Please let me know if I targeted the wrong branch here

qa/config.template.json
```diff
@@ -0,0 +1,9 @@
+{
+  "config": {
+    "languages": {
+      "%s": {
+        "mass_threshold": 20
```
dblandin: Should we be overriding the mass_threshold here?

maxjacobson: Nope, I don't think we need to. Updated

qa/bin/clone_and_test_examples
```diff
@@ -0,0 +1,62 @@
...
+  else
+    LANGS[0]="ruby"
+    LANGS[1]="python"
+    LANGS[2]="php"
+    LANGS[3]="java"
+    LANGS[4]="javascript"
+  fi
+
+  for lang in ${LANGS[*]}; do
+    printf "$(cat "$config_template_path")" "$lang" > "$config_path"
```
wfleming: nice.

PR #213, opened 2017-08-30
Skip unprocessable files

dblandin:
Instead of blowing up, rescue unprocessable files and output a warning.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -6,6 +6,7 @@ Pry.config.pager = false
 Pry.config.color = false
 
 require "cc/logger"
+require "cc/engine/duplication"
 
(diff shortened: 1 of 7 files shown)


spec/spec_helper.rb
```diff
@@ -6,6 +6,7 @@
 Pry.config.color = false
 
 require "cc/logger"
+require "cc/engine/duplication"
```
wfleming: Not clear to me why this was needed here, but there are extant `requires` for this in other specs that can be removed if we're adding it here.

dblandin: Yeah, I ran into a issue when using `RSPEC_ARGS` without this higher level require. Can remove the spec specific ones.

dblandin: Updated.

Issue #206, opened 2017-08-28
s-expression errors


AndreyShpilevoy:
Hi all,

I faced with a very strange issue.
I receive this butch of errors

`ERROR: couldn't find node type SpreadElement in Sexp::NODE_NAMES.
ERROR: couldn't find node type SpreadProperty in Sexp::NODE_NAMES.
ERROR: couldn't find node type YieldExpression in Sexp::NODE_NAMES.
ERROR: couldn't find node type ImportNamespaceSpecifier in Sexp::NODE_NAMES.
ERROR: couldn't find node type JSXEmptyExpression in Sexp::NODE_NAMES.`
And there is no matter what I have in configuration or where it was running (docker or on web site)

Last time I receive this errors with configuration: [Url](https://github.com/AndreyShpilevoy/dem-afterlife/blob/ec68ba0ffe756787c2e3903a5f6a5ba0cd738536/.codeclimate.yml)

toddmazierski:
Hi @AndreyShpilevoy,

Thanks for submitting this issue! I'm happy to help.

I'm having trouble reproducing these errors, however. Using the [Code Climate CLI](https://github.com/codeclimate/codeclimate), I was able to successfully analyze both the commit you linked to (`ec68ba0`) and the `master` branch in this project. Do you mind providing some more information so I can debug further? For example, either:

* a link to a failed build on codeclimate.com
* a commit where the error is reproducible with the Code Climate CLI

AndreyShpilevoy:
yep, it was on branch breadcrumbs
Here is a link to this commit: [https://github.com/AndreyShpilevoy/dem-afterlife/commit/ec68ba0ffe756787c2e3903a5f6a5ba0cd738536](url)

Here is a link to CodeClimate.com build - [https://codeclimate.com/github/AndreyShpilevoy/dem-afterlife/builds/32](url)

Also, interesting, that on github I see that build was failed because of `codeclimate — 382 issues to fix`, but on the build, everything is ok, except exception of course.

One more interesting thing - on Code Climate CLI I had results when duplication engine return for me some valid result, but when I tried to set `dump_ast: true` option to config section - I got the same errors.

toddmazierski:
Thanks, @AndreyShpilevoy!

I now see where the `couldn't find node type` errors are being emitted. My apologies for missing them before. I'm digging into this now to determine the impact and come up with a fix. Please stand by.

[...]

PR #216, opened 2017-08-31
Overwrite Flay's `NODE_NAMES` to suppress ignorable errors (#206)

toddmazierski:
Flay and Code Climate's `CCFlay` have registries of "node names" or "node types" for Ruby, JavaScript, and PHP called `NODE_NAMES`. They look like this:

    {
      alias: 0,
      and: 1,
      …
      kwsplat: 86,
      safe_call: 87
    }

This registry is used in Flay's node hashing function `#pure_ruby_hash`. When a node name being analyzed is missing from the registry, two things happen:

  1. an error is logged: `couldn't find node type`
  2. the node is automatically added to the registry with the next value

For example:

    {
      alias: 0,
      and: 1,
      …
      kwsplat: 86,
      safe_call: 87,
      some_missing_node: 88, # <= automatically added
    }

Since missing nodes are gracefully handled and do not appear to affect analysis accuracy, this commit overwrites `NODE_NAMES.default_proc` to:

  1. "suppress" the error (by leaving it out from the reimplementation)
  2. wrap the mutation in a `Mutex` for thread safety (since `CCFlay` is multi-threaded)

### Tests

Includes a spec to reproduce `couldn't find node type` errors being logged. It failed with:

    expected block to not output to stderr, but output "ERROR: couldn't
    find node type bug_206_node in Sexp::NODE_NAMES.\n"

Add a spec to regression test the automatic assignment of incrementing values for missing nodes.

### DIFF
--- a/lib/ccflay.rb
+++ b/lib/ccflay.rb
@@ -71,6 +71,14 @@ new_nodes.each do |name|
   Sexp::NODE_NAMES[name] = Sexp::NODE_NAMES.size
 end
 
+# Overwrite method from Flay to suppress ignorable errors and for thread safety
+# (please see bug #206)
+node_names_mutex = Mutex.new
+
+Sexp::NODE_NAMES.default_proc = lambda do |hash, key|
+  node_names_mutex.synchronize { hash[key] = Sexp::NODE_NAMES.size }
+end
+
 class Sexp
   attr_writer :mass
 
--- a/spec/cc/ccflay_spec.rb
+++ b/spec/cc/ccflay_spec.rb
@@ -20,4 +20,22 @@ RSpec.describe CCFlay do
       expect(inn.flatter.mass).to eq(8)
     end
   end
+
+  describe Sexp::NODE_NAMES do
+    describe ".default_proc" do
+      it "should automatically assign incrementing values for missing nodes" do
+        node1 = Sexp::NODE_NAMES["some_missing_node1"]
+        node2 = Sexp::NODE_NAMES["some_missing_node2"]
+
+        expect(node1).to be > 0
+        expect(node2).to eq(node1 + 1)
+      end
+
+      context "'couldn't find node type' errors (bug #206)" do
+        it "should suppress them" do
+          expect { Sexp::NODE_NAMES["bug_206_node"] }.to_not output.to_stderr
+        end
+      end
+    end
+  end
 end


lib/ccflay.rb
```diff
@@ -71,6 +71,14 @@ def initialize(option = nil)
   Sexp::NODE_NAMES[name] = Sexp::NODE_NAMES.size
 end
 
+# Overwrite method from Flay to suppress ignorable errors and for thread safety
+# (please see bug #206)
+node_names_mutex = Mutex.new
```
maxjacobson: Is this sufficient? I could be wrong, but I thought that we would need to initialize the mutex before spawning threads so the same mutex is used in each thread

toddmazierski: Great question. Definitely a little odd-looking, but yes, it's sufficient.

This file is unconventional in a few ways, one of them being some code is evaluated when the file _itself_ loaded (as opposed to a `class` or `module` _in_ the file). Here's a [link](https://github.com/codeclimate/codeclimate-duplication/blob/f9dff44d0064168107b4ee1f02214f24dc3daa4c/lib/ccflay.rb#L74-L76) to see it in greater context.

wfleming: Re: Max's point: I think this gets initialized once & only once when the class gets declared, so there's only one instance ever, so it's correct as far as the threading goes.

Stylistically, I find the pattern of local-vars-in-class-decls super weird & a bit of a smell. Could this be a constant instead? Would there be a downside to that stylistic change?

maxjacobson: ah yes the context helps a lot. I was just assuming this was running more than once. But it will be the same mutex each time that lambda is called. Neat, thanks for explaining 👍

toddmazierski: > Could this be a constant instead? Would there be a downside to that stylistic change?

Yup! I think I was just following the lead of the `new_nodes` assignment.

I think the downside is that the mutex, by virtue of being a constant, would be globally scoped, when it's really only needed here.

toddmazierski:
Hello @zenspider,

We were wondering if you had any thoughts on this change, overwriting `NODE_NAMES.default_proc` to:

  1. leave out `couldn't find node type` errors
  2. wrap the mutation in a `Mutex`

Because `new_nodes` is non-exhaustive and we keep expanding our language support, we're seeing this error more often.

More than happy to bring either of these changes upstream to Flay if you'd like. Thanks for your help!

zenspider:
I think that would be fine with one minor exception. Even with the mutex, the parallelization of the processing means that nodes can get added in different order... that might affect hash values, which might affect day to day results or exclusion values/ids or ... something.

Also, it *could* be exhaustive, assuming the languages post specs on their sexps.

toddmazierski:
Thanks, @zenspider.

> Even with the mutex, the parallelization of the processing means that nodes can get added in different order... that might affect hash values, which might affect day to day results or exclusion values/ids

Do you mind elaborating on what effect there might be on exclusions? Node hashes in the same process should be consistent. Are they shared between processes?

zenspider:
I'm just spitballing, not looking at the code.

Everything will be consistent in process which is why I said day-to-day. If the node hash value isn't used for exclusions, then all of this should be a non-issue.

toddmazierski:
Understood. It's a fair point, though. 👍

It might be surprising to find out the nodes values are not consistent from run-to-run. What do you think of this?

```ruby
Sexp::NODE_NAMES = {}

Sexp::NODE_NAMES.default_proc = lambda do |hash, key|
  hash[key] = Zlib.crc32(key.to_s)
end
```

By using a non-cryptographic hash function we could remove the dependency on `new_nodes` (less code and maintenance) and have stable hash values across processes.

I tested this locally and didn't see an impact on accuracy or performance.

toddmazierski:
I've opened up a new pull request (#222) as an alternative to this one with the CRC checksum approach [mentioned above](https://github.com/codeclimate/codeclimate-duplication/pull/216#issuecomment-327305335). I'll be closing this one because I think it's an improvement.

@zenspider et al., your feedback is welcome! Thank you.

PR #217, opened 2017-09-01
Enable QA script on CI

maxjacobson:
I'm deferring enabling it for all of the supported languages, because
we're going to be migrating languages to use the parser backend shortly,
so any errors that may be happening now are less relevant than errors
that occur on the new parser backend.

I'm hopefuly that the `cache_directories` invocation will mean that the
repos will stick around across builds on the same branch, so we can save
a little time by just `git pull`-ing those repos rather than `git
clone`-ing them.

This took about 10 minutes to run locally. It may be different on Circle
CI.

---

**note**: please let me know if this is targeting the wrong branch

### DIFF
--- a/qa/bin/clone_and_test_examples
+++ b/qa/bin/clone_and_test_examples
@@ -18,15 +18,9 @@ if [ -n "$QA" ] || test_on_branch $CIRCLE_BRANCH; then
 
   REPOS_PER_LANGUAGE=20
 
-  if [ -n "$LANGUAGE" ]; then
-    LANGS[0]="$LANGUAGE"
-  else
-    LANGS[0]="ruby"
-    LANGS[1]="python"
-    LANGS[2]="php"
-    LANGS[3]="java"
-    LANGS[4]="javascript"
-  fi
+	if [ -z "$LANGS" ]; then
+		LANGS=(ruby python php java javascript)
+	fi
 
   for lang in ${LANGS[*]}; do
     printf "$(cat "$config_template_path")" "$lang" > "$config_path"
--- a/Makefile
+++ b/Makefile
@@ -13,5 +13,5 @@ citest:
 
 qa: image
 	# requires QA=true to run locally
-	# optionally pass LANGUAGE=x
+	# optionally pass `LANGS="x y"`
 	qa/bin/clone_and_test_examples qa/examples $(IMAGE_NAME)
--- a/circle.yml
+++ b/circle.yml
@@ -14,10 +14,15 @@ dependencies:
       codeclimate/patrick pull || true
     - docker login --username "$DOCKER_USERNAME" --password "$DOCKER_PASSWORD" --email "$DOCKER_EMAIL"
     - make image
+  cache_dependencies:
+    - qa/examples
 
 test:
   override:
     - make citest
+    # N.B. Only QA-ing languages which have been updated to use our new parser
+    # backend
+    - LANGS=java make qa
 
 deployment:
   registry:


toddmazierski:
It seems like this script has surfaced a lot of bugs, @maxjacobson. 👏 

Did you consider using a fixed set of repositories instead of always calling the GitHub API? Relying on a dynamic set of repositories may mean the tests will fail intermittently depending on the result set, or make failures more difficult to reproduce locally.

Perhaps a separate script could be created to update a fixed set of repositories, which can then be checked into source control. Just a thought.

qa/bin/clone_and_test_examples
```diff
@@ -21,11 +21,9 @@ if [ -n "$QA" ] || test_on_branch $CIRCLE_BRANCH; then
...
     LANGS[0]="$LANGUAGE"
   else
-    LANGS[0]="ruby"
-    LANGS[1]="python"
-    LANGS[2]="php"
-    LANGS[3]="java"
-    LANGS[4]="javascript"
+    # N.B. Only QA-ing languages that have been migrated to using our parser
+    # backend
+    LANGS[0]="java"
```
wfleming: This is fine as-is, but seeing the defaults get deleted made me realize we could simplify this approach a good deal & allow overriding the entire array from the outside the script as well.

```
if [ -z "$LANGS" ]; then
  LANGS=(ruby python php java javascript)
fi

for lang in ${LANGS[*]}; do
```

With that formulation, the CI invocation could be `LANGS=java make qa`, and then become `LANGS="java ruby" make qa` to run both, etc. That way instead of having to change the defaults we can hand-pick the ones we want to run at a given time, which seems preferable.

maxjacobson:
Definitely considered that. I was imagining that we would try it this way and then see if we hit that pain and adjust as necessary. I'm not expecting much pain because

1. the list of top repos for a language is probably pretty stable
1. if the list does shift around and surface new errors, that's kind of good - now we're aware of those errors and can fix them. Hopefully this will be rare.
1. I'm hopeful that the output of the script will include sufficient information to identify which file in which repo had the issue. It [prints out](https://github.com/codeclimate/codeclimate-duplication/blob/1c6c650421ce3c5ffe51f1619d8a70ca02a87d01/qa/bin/clone_and_test_examples#L54) the language and github slug

wfleming:
> Did you consider using a fixed set of repositories instead of always calling the GitHub API? Relying on a dynamic set of repositories may mean the tests will fail intermittently depending on the result set, or make failures more difficult to reproduce locally.

I think we consider the first part desirable, and I don't think the second part is actually a problem.

If a new project in a languages becomes extremely popular & happens to use some unusual/new syntax we don't handle well, that project's popularity would probably indicate we'd be more likely to start hitting that syntax in customer code & so we'd want to fix it. So a non-static set of repos makes it easier for us to stay up to date with changing language usage.

And reproduction seems pretty easy: The script logs what it's cloning, and errors generally log where they failed, so reproducing should be a simple matter of cloning the repo from the logs & poking at it locally. If we do run into cases of sometimes-it-errors-sometimes-it-doesn't, we'll need to evaluate how to handle that vis-a-vis CI, but I would still consider those really important to investigate & fix, so if this process happened to turn any of those up, I'd still consider it a net good thing.

qa/bin/clone_and_test_examples
maxjacobson: Oo I like that. Updated.

I don't totally understand where the string is being split into an array, but I did some local tesitng and it does seem to work...

maxjacobson:
Going to merge this and verify on CI that it runs successfully

qa/bin/clone_and_test_examples
wfleming: It's shell word-splitting. Arrays are kinda just some bash fanciness on top of standard word-splitting.

PR #218, opened 2017-09-01
Ignore Import and Package Declarations in Java

ABaldwinHunter:
These lines of code are often considered false positives and can be noisy.

Leans on the new [node filtering](https://github.com/codeclimate/codeclimate-duplication#node-filtering) functionality. While users could exclude these nodes on their own, it seems like a better default behavior to ignore.

Addresses: https://github.com/codeclimate/app/issues/5728

We may want to make similar default exclusions in other languages, such as import statements in JavaScript

Note, I was able to identify the node types we wanted to filter using the `dump_ast` flag.

### DIFF
--- a/lib/cc/engine/analyzers/analyzer_base.rb
+++ b/lib/cc/engine/analyzers/analyzer_base.rb
@@ -51,7 +51,7 @@ module CC
         end
 
         def filters
-          engine_config.filters_for(language)
+          engine_config.filters_for(language) + default_filters
         end
 
         def language
@@ -89,6 +89,10 @@ module CC
 
         def base_points
           self.class::BASE_POINTS
+        end
+
+        def default_filters
+          []
         end
 
         def points_per_overage
--- a/lib/cc/engine/analyzers/analyzer_base.rb
+++ b/lib/cc/engine/analyzers/analyzer_base.rb
@@ -51,7 +51,7 @@ module CC
         end
 
         def filters
-          engine_config.filters_for(language) + default_filters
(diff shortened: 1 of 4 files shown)


lib/cc/engine/analyzers/analyzer_base.rb
```diff
@@ -51,7 +51,7 @@ def files
         end
 
         def filters
-          engine_config.filters_for(language)
+          engine_config.filters_for(language) + default_filters
```
dblandin: What do you think of a union here, so we don't end up processing duplication patterns?

```ruby
engine_config.filters_for(language) | default_filters
```

ABaldwinHunter: great suggestion

dblandin: Actually, I wonder if this works reliably with `Sexp` objects. Realizing that this aren't strings here. Worth checking on.

dblandin: Looks like this works! 👍 

```ruby
irb(main):005:0> a = Sexp::Matcher.parse("(ImportDeclaration ___)")                                                      
=> q(:ImportDeclaration, ___)                                                                                           
irb(main):006:0> b = Sexp::Matcher.parse("(ImportDeclaration ___)")                                                     
=> q(:ImportDeclaration, ___)                                                                                           
irb(main):007:0> [a] | [b]                                                                                              
=> [q(:ImportDeclaration, ___)]
```

PR #219, opened 2017-09-01
Ignore Import and Package Declarations in Java

ABaldwinHunter:
These lines of code are often considered false positives and can be
noisy.

Leans on the new [node
filtering](https://github.com/codeclimate/codeclimate-duplication#node-filtering)
functionality. While users could exclude these nodes on their own, it
seems like a better default behavior to ignore.

Addresses: https://github.com/codeclimate/app/issues/5728

Accidentally opened this against `channel/java` before.

### DIFF
--- a/lib/cc/engine/analyzers/analyzer_base.rb
+++ b/lib/cc/engine/analyzers/analyzer_base.rb
@@ -51,7 +51,7 @@ module CC
         end
 
         def filters
-          engine_config.filters_for(language)
+          engine_config.filters_for(language) | default_filters
         end
 
         def language
@@ -89,6 +89,10 @@ module CC
 
         def base_points
           self.class::BASE_POINTS
+        end
+
+        def default_filters
+          []
         end
 
         def points_per_overage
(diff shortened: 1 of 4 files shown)


lib/cc/engine/analyzers/analyzer_base.rb
```diff
@@ -51,7 +51,7 @@ def files
         end
 
         def filters
-          engine_config.filters_for(language)
+          engine_config.filters_for(language) | default_filters
```
wfleming: I don't know if we want the union of these two sets: a user overriding the default might reasonably want to *not* use one or more of the default filters, yeah?

ABaldwinHunter: I can't imagine a way for a user to override the default, without making it a `default_filters` OR `.codeclimate.yml` backed filters. (since it's an array, instead of hash where we could do some sort of `merge!`)

I'd propose going with this implementation (which seems simplest and desirable for most users), and then changing if we get feedback, because:

I can't imagine users _wanting_ to get `import` and `package` statements flagged. 

I _could_ imagine someone adding a filter for a different node type, then being surprised that it triggered PackageDeclarations and ImportDeclarations to stop being ignored.

The (IMO less desirable) alternative would be to use an `OR` and update documentation, see if that generates confusion.

WDYT?

dblandin: I think this behavior sounds fine as a starting point. Any user-supplied filters would be added to our default list, which we've determined safe to ignore for all users.

If we made it an override, users would have to pick out filters from the default set that they'd like to ignore, which seems undesirable.

PR #222, opened 2017-09-07
Populate Flay's `NODE_NAMES` on-demand with CRC checksum (#206)

toddmazierski:
## Background

This is an alternative to the solution originally proposed in #216.

## Summary

Overwrite `NODE_NAMES` from Flay to assign all values on-demand instead of using a predefined registry. Use CRC checksums so hash values are order-independent (i.e. consistent between runs).

As a side-effect (but the motivation behind this change), this fixes an error from Flay: `couldn't find node type`. This is being emitted more often now that we're expanding our language support.

Populating values on-demand is preferable to simply updating the registry since it's less work to maintain, less code, and there's virtually no performance penalty.

## Testing

Using the [Code Climate CLI](https://github.com/codeclimate/codeclimate), averages of 9 runs of duplication analysis of https://github.com/rails/rails/commit/806f96ceaecbf9cf8d2fdbbdaf1fb2a499769cbf were:

* `master`: 65.96s
* `206-node-type-errors-crc` (this branch): 65.93s

The analysis results were identical across all runs (2,308 issues found).

### DIFF
--- a/lib/ccflay.rb
+++ b/lib/ccflay.rb
@@ -3,6 +3,7 @@
 require "flay"
 require "concurrent"
 require "digest"
+require "zlib"
 
 ##
 # A thread-safe and stable hash subclass of Flay.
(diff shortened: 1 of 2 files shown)


wfleming:
@toddmazierski after you ship this we should get `channel/cronopio` rebased on `master` so we can get the same change in quality model.

PR #230, opened 2017-09-21
Change reported check name to match QM formatting

wfleming:
The config for QM uses dasherized lowercase (`identical-code`), whereas
the reported check names here were more spoken language (`Identical
code`). The change here changes reported check names to match what QM
says & is configured via moving forward.

Note this was intentionally written to *NOT* change fingerprints. The
old style check names are still calculated and used for those.

This also keeps compatability with an older style of configuring
checks. e.g.

```
engines:
  duplication:
    checks:
      Similar code:
        enabled: false
```

That style of config is deprecated, for QM checks/engines, but the
implementation here will still use that config if present, unless
it's overridden by the newer QM style.

### DIFF
--- a/lib/cc/engine/analyzers/analyzer_base.rb
+++ b/lib/cc/engine/analyzers/analyzer_base.rb
@@ -71,7 +71,7 @@ module CC
         end
 
         def calculate_points(violation)
-          overage = violation.mass - check_mass_threshold(violation.inner_check_name)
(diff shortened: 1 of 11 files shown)


wfleming:
@maxjacobson After further thought I decided to keep support for the old style: we're still going to support that style for non-QM engines, so we as well take the extra step here to keep old configs working. I requested a re-review.

PR #234, opened 2017-09-25
Bump cc-parser to b378

maxjacobson:
related to https://github.com/codeclimate/app/issues/5902

These tests were failing before we bumped the parser image to include these changes: https://github.com/codeclimate/codeclimate-parser/pull/89

paired with @larkinscott on this

### DIFF
--- a/spec/cc/engine/analyzers/java/java_spec.rb
+++ b/spec/cc/engine/analyzers/java/java_spec.rb
@@ -4,7 +4,7 @@ require "cc/engine/analyzers/engine_config"
 require "cc/engine/analyzers/file_list"
 
 module CC::Engine::Analyzers
-  RSpec.describe Java::Main, in_tmpdir: true, focus: true do
+  RSpec.describe Java::Main, in_tmpdir: true do
     include AnalyzerSpecHelpers
 
     describe "#run" do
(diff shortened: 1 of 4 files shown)


spec/cc/engine/analyzers/java/java_spec.rb
```diff
@@ -142,6 +142,76 @@ module CC::Engine::Analyzers
         issues = run_engine(engine_conf).strip.split("\0")
         expect(issues).to be_empty
       end
+
+      it "prints an issue for similar code when the only difference is the value of two hexadecimal numbers" do
```
ABaldwinHunter: I like this test a lot, but what do you think about calling this something a bit more generic and simple, like:

`it "prints an issue for similar code" do`

(that would have failed before too right?)

spec/cc/engine/processed_source_spec.rb
```diff
@@ -0,0 +1,39 @@
+require "spec_helper"
+require "cc/engine/processed_source"
+
+RSpec.describe CC::Engine::ProcessedSource, in_tmpdir: true do
+  include AnalyzerSpecHelpers
+
+  describe "#ast" do
```
ABaldwinHunter: I like having tests for this class since these are public methods!

It does feel a little like we're testing the parser here, but I'm okay with it - since this code seems unlikely to change frequently - and kind of serves as a stand in integration test. Also, it would have been helpful in this case.

PR #238, opened 2017-09-27
[WIP] Add TypeScript support

brynary:
Parser image needs to be bumped further after the latest TypeScript parser is merged.

This seems to be close, but I'm getting some odd failures in the specs, wherein the output results don't match what I'd expect. Specifically:

* The line numbers reported in the `other_locations` data in the spec for similar code look wrong
* The identical code spec is reporting as similar, not identical

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM codeclimate/codeclimate-parser:b378
+FROM codeclimate/codeclimate-parser:b384
 LABEL maintainer="Code Climate <hello@codeclimate.com>"
 
 # Reset from base image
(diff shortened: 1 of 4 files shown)


brynary:
Blocked on the Parser fixes that were identified in #239 and need to be fixed first

andrew-matteson:
Since #239 looks merged, what's the status on this?

If there was a simple path forward, I'd be happy to try to help. :-)

brynary:
@andrew-matteson hey! Sorry for the late reply here... this was handled via another PR and is all set now! Thanks for the offer!

PR #239, opened 2017-09-27
Move JavaScript to parser backend

maxjacobson:
Hitting a few interesting hurdles so far. I think I may need to make some parser tweaks before this will be workable. Just opening for now to make it easier to find tomorrow 😉.


brynary:
@maxjacobson would be interested to hear about these hurdles! (Context: I am trying to get TypeScript working in here, and am probably running into some of the same ones)

maxjacobson:
@brynary the issue is that we're getting incorrect line numbers, due to a confluence fo events:

1.  the AST produced by the parser contains `extra` nodes (we're actually tracking that [here](https://github.com/codeclimate/app/issues/5829)) which don't have a type or a location. The parser is giving these a default location of `1`, and a type of `_unknown`. It adds these to, for example, string literals
1. And then, here in duplication, when determining the location for a duplicative node, [we find the lowest start value for a node and all of its children](https://github.com/codeclimate/codeclimate-duplication/blob/4ca4466925f1c5af5b0643a5f9bf05241f95f5df/lib/cc/engine/analyzers/sexp_lines.rb#L28-L31) (apparently to work around a bug in the ruby parser)
1. Because our string literals contain a child node, extra, with a start value of  `1` (from the defaulting), we think the whole node starts on line 1

This didn't previously affect us because we were [scrubbing out](https://github.com/codeclimate/codeclimate-duplication/blob/4ca4466925f1c5af5b0643a5f9bf05241f95f5df/lib/cc/engine/analyzers/javascript/parser.js#L13-L14) the values from extra nodes, but that behavior didn't carry over to the parser (perhaps we didn't know why we had it, and optimistically removed it?)

The plan to continue is to update the parser to flatten `extra` nodes into their parents, so `raw` and `rawValue` become properties fo the string literal, which has a proper location, rather than a locationless child

brynary:
Ahhhh... this is definitely affecting TypeScript. Great catch!

maxjacobson:
This is green with these parser changes: https://github.com/codeclimate/codeclimate-parser/pull/93

going to push that through, then update this PR and open it for review

maxjacobson:
I've done some local QA on our app repo to compare the results. Here are some findings:

- a few line numbers changed, but in ways that I think are more correct (instead of highlighting [this](https://github.com/codeclimate/app/blob/bb59e47f52266cc3bac24b3e400399ab46a1ea88/app/assets/javascripts/redesign/modules/repo_key.js#L76-L88, it highlights [this](https://github.com/codeclimate/app/blob/bb59e47f52266cc3bac24b3e400399ab46a1ea88/app/assets/javascripts/redesign/modules/repo_key.js#L84-L88) as being similar to [this](https://github.com/codeclimate/app/blob/bb59e47f52266cc3bac24b3e400399ab46a1ea88/app/assets/javascripts/redesign/modules/add_repo.js#L114-L118), for example)
- some new issues are being found that weren't previously being found (for example [this](https://github.com/codeclimate/app/blob/bb59e47f52266cc3bac24b3e400399ab46a1ea88/app/assets/javascripts/poll_scheduler.js#L27-L33) is indeed similar to [this](https://github.com/codeclimate/app/blob/bb59e47f52266cc3bac24b3e400399ab46a1ea88/app/assets/javascripts/redesign/modules/poll_scheduler.js#L37-L43))

Other than that, it's pretty much the same.

Therefore, I feel good about this change. Going to open up for review now.

lib/cc/engine/analyzers/analyzer_base.rb
```diff
@@ -120,6 +123,35 @@ def file_list
...
+            CC.logger.warn(ex.message)
+            CC.logger.debug { "Contents:\n#{processed_source.raw_source}" }
+            nil
+          else
+            CC.logger.error("Error processing file: #{processed_source.path}")
+            CC.logger.error(ex.message)
+            CC.logger.debug { "Contents:\n#{processed_source.raw_source}" }
+            raise
+          end
+        end
```
toddmazierski: I see this code is just being moved from one file to the other, but one potential improvement is to define multiple `rescue` statements, rather than `if` and `#is_a?`. For example:

```ruby
def parse(file, request_path)
  // …
rescue CC::Parser::Client::HTTPError => ex
  // …
rescue CC::Parser::Client::NestingDepthError => ex
  // …
else
  // …
end
```

lib/cc/engine/analyzers/java/main.rb
```diff
@@ -20,43 +18,16 @@ class Main < CC::Engine::Analyzers::Base
...
-              CC.logger.warn(ex.message)
-              CC.logger.debug { "Contents:\n#{processed_source.raw_source}" }
-              nil
-            else
-              CC.logger.error("Error processing file: #{processed_source.path}")
-              CC.logger.error(ex.message)
-              CC.logger.debug { "Contents:\n#{processed_source.raw_source}" }
-              raise
-            end
+            parse(file, "/java")
```
toddmazierski: Out of curiosity, why not continue to use a constant?

PR #240, opened 2017-09-28
Modernize README

wfleming:
- Use the newer config keys/formats
- Remove example languages config: this is no longer the preferred usage
- Remove engine-level exclude pattern docs: this is a global feature,
and shouldn't be documented here, as it'll just fall out date.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,25 +21,6 @@ simple. We have a [docs page][what-is-duplication] explaining the algorithm.
 
 ## Configuring
 
-### Languages
-
-By enabling the duplication engine with the Code Climate CLI, all supported
-languages are configured by default, but we recommend adjusting this
-configuration to enable only the languages you care about. If you have a project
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -73,7 +52,7 @@ the Hash type to support extra configuration.
 By default, the duplication engine will report code that has been duplicated in just two locations. You can be less strict by only raising a warning if code is duplicated in three or more locations only. To adjust this setting, add a `count_threshold` key to your config. For instance, to use the default `mass_threshold` for ruby, but to enforce the [Rule of Three][rule-of-three], you could use this configuration:
 
 ```yaml
-engines:
+plugins:
   duplication:
     enabled: true
     config:
```
dblandin: We we properly merge the plugin-style configuration with the similar/identical check configuration?

wfleming: For the purposes of legacy config support, yes.

PR #241, opened 2017-09-28
Remove qa scripts

wfleming:
These live in https://github.com/codeclimate/qm_qa now


maxjacobson:
There are some lines you can remove from gitignore and dockerignore now

wfleming:
@maxjacobson thanks, removed.

maxjacobson:
OK since I already drove by commented I'll drive-by approve :)

DEVELOPERS.md
```diff
@@ -8,3 +8,10 @@
 * install composer: `curl "https://getcomposer.org/installer" | php`
 * update `composer.lock`: `php composer.phar update`
 * `rm composer.phar`
+
+## QA
+
+There is an automated [QA tool](https://github.com/codeclimate/qm_qa) that can
```
chrishulton: my only question is around linking to this here, since its a private repo, and duplication is public.

wfleming: Boo, good point. I'm inclined to leave it, since QM stuff will be open sourced at some point, and this is an important and valuable step for people adding QM language support. If it comes up as a problem with an open source contributor before we open source it, we'll address it then.

PR #245, opened 2017-10-09
Ignore PHP namespace/use statements

brynary:


wfleming:
Good point, @wilson, we are using a new enough Ruby here now so I updated that. I will merge on green.

PR #248, opened 2017-10-10
Ignore import statements for JS duplication

brynary:
Happy


lib/cc/engine/analyzers/javascript/main.rb
```diff
@@ -13,6 +13,9 @@ class Main < CC::Engine::Analyzers::Base
           ].freeze
           LANGUAGE = "javascript"
           DEFAULT_MASS_THRESHOLD = 40
+          DEFAULT_FILTERS = [
```
ABaldwinHunter: nit: alphabetize this constant?

PR #250, opened 2017-10-11
Ignore variable declarations in JS

brynary:
Erring on the side of less is more when it comes to duplication, this avoids us flagging `require()`s as duplication.


wfleming:
This seems overly broad. Wouldn't excluding only assignments from calls to `require` be more precise, and also more similar to the approach taken for the imports? It seems like blocks of var decls in other contexts probably should be fair game for the analysis.

The pattern to do it that way is:

```
(VariableDeclarator _ (init (CallExpression (_ (Identifier require)) ___)))
```

brynary:
Good point. Updated.

PR #256, opened 2017-10-18
Integrate PHP parser server

toddmazierski:
Replace all the existing PHP parsing code with an integration of the PHP parser server.

There are some major differences between the old and new parsers worth noting:

1. The AST is now much deeper overall. For example, an `else` branch from a unit test case has gone from 4 to 8 levels of depth. As a result, the masses have increased across the board, invalidating all `fingerprint` values and necessitating a change to the `DEFAULT_MASS_THRESHOLD` to keep the number of issues roughly consistent.

2. The AST now includes `comments` nodes, which need to be filtered out to avoid having them contribute to analysis.

3. The AST now has the correct `end` position for functions (the closing curly brace).

4. The new name for `use` statement nodes is `Stmt_Use`.

Task: codeclimate/app#5898.

## TODO

- [x] I've been doing a lot of testing with the [Symfony](https://github.com/symfony/symfony) project to calibrate the `DEFAULT_MASS_THRESHOLD` and find bugs. While this is being reviewed, I'm going to find and test with a few more.


toddmazierski:
It looks like getting the threshold right is a little more art than science. I collected data from a handful of popular PHP respositories here for a few different mass thresholds [in this spreadsheet](https://docs.google.com/a/codeclimate.com/spreadsheets/d/1wE5f4J4g0KB3LNu2icO_HLEyBLnEErnsWJxr83EH0S0/edit?usp=sharing). I'm thinking a value of `75` is the sweet spot, but please let me know if you have any thoughts.

screen shot 2017-10-18 at 6 21 17 pm

wfleming:
I'm asking Devon to review instead of me if he has time: he helped break ground on switching these languages to the parser, and I don't think I have time to effectively review this today.

lib/cc/engine/analyzers/php/main.rb
```diff
@@ -14,31 +11,31 @@ class Main < CC::Engine::Analyzers::Base
...
           ].freeze
-          DEFAULT_MASS_THRESHOLD = 28
+          DEFAULT_MASS_THRESHOLD = 75
           DEFAULT_FILTERS = [
-            "(use ___)".freeze,
+            "(Stmt_Use ___)",
           ].freeze
           POINTS_PER_OVERAGE = 100_000
+          REQUEST_PATH = "/php"
+          COMMENT_MATCHER = Sexp::Matcher.parse("(_ (comments ___) ___)")
```
dblandin: Is is possible to add this to `DEFAULT_FILTERS` instead of deleting the nodes?

spec/cc/engine/analyzers/php/main_spec.rb
```diff
@@ -38,14 +38,14 @@
       expect(json["categories"]).to eq(["Duplication"])
       expect(json["location"]).to eq({
         "path" => "foo.php",
-        "lines" => { "begin" => 2, "end" => 6 },
+        "lines" => { "begin" => 2, "end" => 8 },
       })
-      expect(json["remediation_points"]).to eq(900_000)
+      expect(json["remediation_points"]).to eq(2_200_000)
```
dblandin: Should we adjust our remediation point multiplier as well so that the same duplication issues are given a similar remediation estimate?

lib/cc/engine/analyzers/php/ast.rb
```diff
@@ -1,51 +0,0 @@
-# frozen_string_literal: true
-
-require "cc/engine/analyzers/php/nodes"
-
-module CC
-  module Engine
-    module Analyzers
-      module Php
```
dblandin: I thnk we usually capitalize acronym class/module names. So maybe this should be `PHP`?

lib/cc/engine/analyzers/php/main.rb
toddmazierski: Unfortunately, it's not. That's the first thing I tried! Adding this to `DEFAULT_FILTERS` causes the entire function to be omitted from analysis.

spec/cc/engine/analyzers/php/main_spec.rb
toddmazierski: Ah, I didn't know there was a multiplier to adjust. I'll take a look at that. 👍

lib/cc/engine/analyzers/php/main.rb
dblandin: Hmmm, what about `(comments ___)` as a filter? Would that match just the comment instead of the entire function?

toddmazierski:
@dblandin, I've adjusted the `POINTS_PER_OVERAGE` from `100_000` to `40_000`, again, trying to strike a balance between a couple projects. It's _extremely_ rough, but we're closer than with the previous value. The values within the test case are also much closer.

### Symfony

#### Cronopio

260,400,000 total points
152 total issues
1,713,157 points per issue

#### This branch (40,000 points per overage)

318,800,000 total points (+58,400,000)
210 total issues (+58)
1,518,095 points per issue (-195,062)

###  Codeigniter

#### Cronopio

565,200,000 total points
184 total issues
3,071,739 points per issue

#### This branch (40,000 points per overage)

594,360,000 total points (-29,160,000)
148 total issues (-36)
4,015,945 points per issue (+944,206)

spec/cc/engine/analyzers/php/main_spec.rb
toddmazierski: Please see [here](https://github.com/codeclimate/codeclimate-duplication/pull/256#issuecomment-338010509) for response!

lib/cc/engine/analyzers/php/main.rb
toddmazierski: It doesn't, please see the test output below.

I discussed this with @wfleming, too, and he said that while this is not entirely desirable behavior, this is how we originally intended filtering to work.

```

Failures:

  1) CC::Engine::Analyzers::Php::Main#run comments ignores PHPDoc comments
     Failure/Error: expect(issues.length).to be > 0
       expected: > 0
            got:   0
     # ./spec/cc/engine/analyzers/php/main_spec.rb:232:in `block (4 levels) in <top (required)>'
     # ./spec/spec_helper.rb:27:in `block (4 levels) in <top (required)>'
     # ./spec/spec_helper.rb:26:in `chdir'
     # ./spec/spec_helper.rb:26:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:23:in `block (2 levels) in <top (required)>'

  2) CC::Engine::Analyzers::Php::Main#run comments ignores one-line comments
     Failure/Error: expect(issues.length).to be > 0
       expected: > 0
            got:   0
     # ./spec/cc/engine/analyzers/php/main_spec.rb:266:in `block (4 levels) in <top (required)>'
     # ./spec/spec_helper.rb:27:in `block (4 levels) in <top (required)>'
     # ./spec/spec_helper.rb:26:in `chdir'
     # ./spec/spec_helper.rb:26:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:23:in `block (2 levels) in <top (required)>'
```

wfleming: > while this is not entirely desirable behavior, this is how we originally intended filtering to work

Well,. it's how the *author* of the filtering intended it to work, clearly. I'm not sure it's how "we" as in the product team intended it to work. I honestly don't know if if was what we intended or it was miscommunicated requirements & not appropriately QAed.

toddmazierski:
Hi! I've created a [new tab on the spreadsheet](https://docs.google.com/spreadsheets/d/1wE5f4J4g0KB3LNu2icO_HLEyBLnEErnsWJxr83EH0S0/edit#gid=1258698183) I've been using for analysis to expand upon the `POINTS_PER_OVERAGE` determination started [in this earlier comment](https://github.com/codeclimate/codeclimate-duplication/pull/256#issuecomment-338010509). With a few new projects added, it seems like `35_000` points (from `40_000` in our first pass) may be a better choice. Please let me know what you think!

dblandin:
`35_000` sounds good to me 👍

PR #257, opened 2017-10-19
Fix bug in parse rescue && refactor

ABaldwinHunter:
Addresses: https://github.com/codeclimate/app/issues/6223

Fixes bug and adds a spec.

I also refactored the error handling to avoid nested re-raising in
multiple places - which led to some confusion in debugging new tests.

### DIFF
--- a/lib/cc/engine/analyzers/analyzer_base.rb
+++ b/lib/cc/engine/analyzers/analyzer_base.rb
@@ -126,31 +126,28 @@ module CC
 
         def parse(file, request_path)
           processed_source = ProcessedSource.new(file, request_path)
-
           SexpBuilder.new(processed_source.ast, file).build
-        rescue CC::Parser::Client::HTTPError => ex
-          unexpected_parse_error!(ex) unless ex.response_status.to_s.start_with?("4")
-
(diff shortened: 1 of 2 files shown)


lib/cc/engine/analyzers/analyzer_base.rb
```diff
@@ -126,31 +126,28 @@ def skip?(_path)
...
-          nil
         rescue => ex
-          unexpected_parse_error!(ex)
+          handle_exception(processed_source, ex)
         end
 
-        def unexpected_parse_error!(ex)
-          CC.logger.error("Error processing file: #{processed_source.path}")
-          CC.logger.error(ex.message)
+        def handle_exception(processed_source, ex)
```
maxjacobson: I do recall having a conversation around method naming, and @toddmazierski not loving "handle" for a method that sometimes re-raises. If you find that surprising, consider considering another name.

ABaldwinHunter: interesting. I also considered something more like:

```
rescue => ex
  log_exception(ex)
  if reraise_exception?(ex)
    raise ex
  end
end

def log_exception(processed_source, ex)
end

def reraise_exception?(ex)
end
```
@toddmazierski any preference?

PR #258, opened 2017-10-23
Only use `SexpLines` for old parsers

toddmazierski:
Addresses codeclimate/app#6227.

### DIFF
--- a/lib/cc/engine/analyzers/analyzer_base.rb
+++ b/lib/cc/engine/analyzers/analyzer_base.rb
@@ -90,6 +90,11 @@ module CC
           sexp
         end
 
+        # Please see: codeclimate/app#6227
+        def use_sexp_lines?
+          true
+        end
+
         private
 
         attr_reader :engine_config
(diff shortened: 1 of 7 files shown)


toddmazierski:
A before and after from the [repository mentioned in the escalation](https://github.com/gramps-graphql/data-source-imdbapi):

screen shot 2017-10-23 at 11 57 37 am

I've done the same on [facebook/react](https://github.com/facebook/react) and the results look similarly good:

screen shot 2017-10-23 at 11 54 35 am

toddmazierski:
I've also tested with [Symfony](https://github.com/symfony/symfony) and several Java projects from Square. No changes here—the output is identical.

toddmazierski:
A couple screenshots [from the UI](https://codeclimate.com/github/codeclimate-testing/react/issues?category=duplication) of the bugfix, too:

### Before

screen shot 2017-10-23 at 12 37 09 pm

### After

screen shot 2017-10-23 at 12 44 52 pm

PR #274, opened 2017-11-14
Replace invalid UTF-8 characters in AST values

toddmazierski:
Addresses codeclimate/escalations#180.

Add a test and a fixture to reproduce the `EncodingError: invalid
encoding symbol`.

Source of the fixture: https://github.com/html5lib/html5lib-python

### DIFF
--- a/lib/cc/engine/analyzers/node.rb
+++ b/lib/cc/engine/analyzers/node.rb
@@ -47,7 +47,7 @@ module CC
             elsif value.is_a?(Hash)
               create_sexp(key.to_sym, self.class.new(value, @file, @line).format)
             else
-              value.to_s.to_sym
+              replace_invalid_characters(value.to_s).to_sym
             end
           end
         end
@@ -65,6 +65,14 @@ module CC
 
         def line_number
           raise "Subclass must implement `line_number`"
+        end
+
+        def replace_invalid_characters(value)
+          value.encode(Encoding::UTF_8,
+            invalid: :replace,
+            undef: :replace,
+            replace: "?",
+          )
         end
       end
     end
--- a/lib/cc/engine/analyzers/node.rb
+++ b/lib/cc/engine/analyzers/node.rb
@@ -47,7 +47,7 @@ module CC
             elsif value.is_a?(Hash)
               create_sexp(key.to_sym, self.class.new(value, @file, @line).format)
             else
-              replace_invalid_characters(value.to_s).to_sym
+              value.to_s.to_sym
             end
           end
         end
@@ -65,14 +65,6 @@ module CC
 
         def line_number
           raise "Subclass must implement `line_number`"
-        end
-
-        def replace_invalid_characters(value)
-          value.encode(Encoding::UTF_8,
-            invalid: :replace,
-            undef: :replace,
-            replace: "?",
-          )
         end
       end
     end
--- a/lib/cc/engine/analyzers/python/parser.py
+++ b/lib/cc/engine/analyzers/python/parser.py
@@ -42,4 +42,4 @@ if __name__ == '__main__':
     source = ""
     for line in sys.stdin.readlines():
         source += line
-    print(json.dumps(to_json(ast.parse(source))))
+    print(json.dumps(to_json(ast.parse(source)), encoding='unicode_escape'))
--- a/spec/cc/engine/analyzers/python/main_spec.rb
+++ b/spec/cc/engine/analyzers/python/main_spec.rb
@@ -135,6 +135,17 @@ def a(thing):
     expect(run_engine(engine_conf)).to eq("")
   end
 
+  context "EncodingError: invalid encoding symbol (codeclimate/escalations#180)" do
+    it "parses" do
+      path = fixture_path("escalation-180.py")
+      create_source_file("foo.py", File.read(path))
+      create_source_file("bar.py", File.read(path))
+
+      issues = run_engine(engine_conf).strip.split("\0")
+      expect(issues).to_not be_empty
+    end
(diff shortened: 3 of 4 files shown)


wfleming:
We don't support non-UTF-8 files. This also seems like a temporary patch: when Python gets switched over to parser infrastructure here, I think we're going to run into this all over again.

If this file legitimately is not UFT-8, we should probably tell the customer we don't support non-UTF-8 and exclude it from their analysis via exclude patterns.

ABaldwinHunter:
> If this file legitimately is not UFT-8, we should probably tell the customer we don't support non-UTF-8 and exclude it from their analysis via exclude patterns.

I guess that's consistent with what we do in the structure engine right?

wfleming:
I don't know if we've hade CX inbounds related to encodings on structure recently that have led to this, though I'd be interested in knowing if this file does analyze in structure & why it does if it does.

Note that with some recent parser client changes in https://github.com/codeclimate/codeclimate-parser/pull/128, structure will probably start explicitly erroring on this file for an invalid encoding. If we are going to take the "best effort to translate to UTF-8" approach, we'll need to do that in the parser client as well.

toddmazierski:
Agreed! This is totally a patch to prevent a hard crash. We can also add this exception to [`RESCUABLE_ERRORS`](https://github.com/codeclimate/codeclimate-duplication/blob/e5e28c4ef3e00b261352de27ea89b81bafa403c1/lib/cc/engine/analyzers/analyzer_base.rb#L10-L16) if that's a better one.

Here is the specific `value` in the fixture that causes the exception:

```ruby
value
// => "[\u0001-\b\v\u000E-\u001F\u007F-\u009F?\xED\xBF\xBF\uFDD0-\uFDEF\uFFFE\uFFFF\u{1FFFE}\u{1FFFF}\u{2FFFE}\u{2FFFF}\u{3FFFE}\u{3FFFF}\u{4FFFE}\u{4FFFF}\u{5FFFE}\u{5FFFF}\u{6FFFE}\u{6FFFF}\u{7FFFE}\u{7FFFF}\u{8FFFE}\u{8FFFF}\u{9FFFE}\u{9FFFF}\u{AFFFE}\u{AFFFF}\u{BFFFE}\u{BFFFF}\u{CFFFE}\u{CFFFF}\u{DFFFE}\u{DFFFF}\u{EFFFE}\u{EFFFF}\u{FFFFE}\u{FFFFF}\u{10FFFE}\u{10FFFF}]"

value.encoding
// => #<Encoding:UTF-8>

value.to_sym
// => EncodingError: invalid encoding symbol

replace_invalid_characters(value)
// => "[\u0001-\b\v\u000E-\u001F\u007F-\u009F???-???\uFDD0-\uFDEF\uFFFE\uFFFFU0001FFFEU0001FFFFU0002FFFEU0002FFFFU0003FFFEU0003FFFFU0004FFFEU0004FFFFU0005FFFEU0005FFFFU0006FFFEU0006FFFFU0007FFFEU0007FFFFU0008FFFEU0008FFFFU0009FFFEU0009FFFFU000AFFFEU000AFFFFU000BFFFEU000BFFFFU000CFFFEU000CFFFFU000DFFFEU000DFFFFU000EFFFEU000EFFFFU000FFFFEU000FFFFFU0010FFFEU0010FFFF]"
```

Is it possible that the file is properly encoded, but `invalid_unicode_re` _represents_ invalid characters?

wfleming:
Good question. From some quick poking, the file itself does appear to be UTF-8:

```
[codeclimate-duplication]  iconv -f UTF-8 -t UTF-8 foo.py > /dev/null                    shipbot/bump-parser-b577
[codeclimate-duplication]  echo $?                                                       shipbot/bump-parser-b577
0
[codeclimate-duplication]  irb                                                           shipbot/bump-parser-b577
irb(main):001:0> s = File.read("foo.py"); nil
=> nil
irb(main):002:0> s.encoding
=> #<Encoding:UTF-8>
irb(main):003:0> s.valid_encoding?
=> true
irb(main):004:0>
```

(foo.py is what I downloaded your fixture to locally).

This leads me back around to basically changing my mind on my initial reaction here. We've made other, similar changes to the parser in other places (e.g. https://github.com/codeclimate/codeclimate-parser/pull/105) to keep the output of the python parser representable. So this does seem like another frustrating case where the you give the Python parser valid UTF-8, and it actually *interprets* escape sequences so the output AST might *not* be valid UTF-8.

If/when we can get the rest of duplication onto the HTTP parser, the parser will take care of this concern for us, but now that it turns out that this is a valid UTF-8 file effectively being mangled by the Python parser itself, I regret my initial reaction here, and I think your initial idea of doing whatever you can to re-encode the AST to be UTF-8 was the right one and we should move forward with it. My only remaining question would be whether, instead of doing it at the Ruby layer, we should consider back-porting the relevant aspects of the change in https://github.com/codeclimate/codeclimate-parser/pull/105 to the Python parser here if that will also fix it.

toddmazierski:
> …instead of doing it at the Ruby layer, we should consider back-porting the relevant aspects of the change in codeclimate/codeclimate-parser#105 to the Python parser here if that will also fix it.

I'll give that a try! The test should confirm whether or not it worked.

toddmazierski:
@wfleming, I ported over codeclimate/codeclimate-parser#105 in e1b730fee1d16e07f2fc8e242a8dba92e36e8cc2. I got the original `EncodingError: invalid encoding symbol` and a couple of new failed cases.

toddmazierski:
We were able to work around the problem by other means, and this patch might cause more trouble than its worth. Closing this for now!

PR #275, opened 2017-11-14
Logs file name and re-raises exceptions during processing

chrishulton:
The engine can end up crashing with cryptic errors depending on the
source being analyzed, such as `stack level too deep
(SystemStackError)`.

This rescues exceptions encountered during file processing, logs the
file name, and then re-raises the exception. This will provide more
insight into the issue, and allows users to take action if appropriate
(for instance, excluding the file).

related: https://github.com/codeclimate/escalations/issues/179

### DIFF
--- a/lib/cc/engine/analyzers/reporter.rb
+++ b/lib/cc/engine/analyzers/reporter.rb
@@ -72,13 +72,18 @@ module CC
           processed_files_count = Concurrent::AtomicFixnum.new
 
           pool.run do |file|
-            CC.logger.debug("Processing #{lang} file: #{file}")
+            begin
+              CC.logger.debug("Processing #{lang} file: #{file}")
 
-            sexp = language_strategy.run(file)
+              sexp = language_strategy.run(file)
 
-            process_sexp(sexp)
+              process_sexp(sexp)
 
-            processed_files_count.increment
+              processed_files_count.increment
+            rescue Exception => ex
+              CC.logger.warn("Error processing file: #{file}")
+              raise ex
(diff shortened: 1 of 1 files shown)


lib/cc/engine/analyzers/reporter.rb
```diff
@@ -72,13 +72,18 @@ def process_files
...
-            sexp = language_strategy.run(file)
+              sexp = language_strategy.run(file)
 
-            process_sexp(sexp)
+              process_sexp(sexp)
 
-            processed_files_count.increment
+              processed_files_count.increment
+            rescue Exception => ex
+              CC.logger.warn("Error processing file: #{file}")
```
chrishulton: there is no existing spec code for the reporter, IMO this change is small enough where it's not worth setting that up.

i did verify this with a problematic repo locally:
```
codeclimate-parser socket not present
waiting 1s...
W, [2017-11-14T22:22:03.709182 #1]  WARN -- : Error processing file: path/to/file.rb
/var/lib/gems/2.3.0/gems/sexp_processor-4.10.0/lib/sexp.rb:57:in `new': stack level too deep (SystemStackError)
```

chrishulton:
@nporteschaikin I felt a little weird rescuing `Exception` as well, but yeah since we are re-raising it, I think it should be ok.  Also, this particular exception `SystemStackError` looks to just be a descendant of `Exception`, rather than `StandardError`.

```
[3] pry(main)> SystemStackError.superclass
=> Exception
```

PR #288, opened 2017-11-30
Add Go duplication checks to channel/go-beta

larkinscott:
- Tested locally, pushing up to channel/go-beta for further testing before we release
- Addresses https://github.com/codeclimate/app/issues/6369

### DIFF
--- a/lib/cc/engine/analyzers/go/main.rb
+++ b/lib/cc/engine/analyzers/go/main.rb
@@ -12,7 +12,7 @@ module CC
         class Main < CC::Engine::Analyzers::Base
           LANGUAGE = "go".freeze
           PATTERNS = ["**/*.go"].freeze
-          DEFAULT_MASS_THRESHOLD = 25
+          DEFAULT_MASS_THRESHOLD = 30
           DEFAULT_FILTERS = [
             "(ImportSpec ___)".freeze,
             "(File ___)".freeze,
--- a/lib/cc/engine/analyzers/go/main.rb
+++ b/lib/cc/engine/analyzers/go/main.rb
@@ -10,19 +10,22 @@ module CC
     module Analyzers
       module Go
         class Main < CC::Engine::Analyzers::Base
-          LANGUAGE = "go".freeze
+          LANGUAGE = "go"
           PATTERNS = ["**/*.go"].freeze
           DEFAULT_MASS_THRESHOLD = 30
           DEFAULT_FILTERS = [
-            "(ImportSpec ___)".freeze,
-            "(File ___)".freeze,
-            "(Comment ___)".freeze,
+            "(ImportSpec ___)",
           ].freeze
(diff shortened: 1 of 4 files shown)


lib/cc/engine/analyzers/go/main.rb
```diff
@@ -0,0 +1,41 @@
...
+require "json"
+require "cc/engine/analyzers/reporter"
+require "cc/engine/analyzers/analyzer_base"
+
+module CC
+  module Engine
+    module Analyzers
+      module Go
+        class Main < CC::Engine::Analyzers::Base
+          LANGUAGE = "go".freeze
```
toddmazierski: Most of these `#freeze` calls are redundant because of the magic `frozen_string_literal` comment at the top of the file.

spec/cc/engine/analyzers/go/main_spec.rb
```diff
@@ -0,0 +1,180 @@
...
+          /* This is a also multiline comment */
+          /* This is a also multiline comment */
+
+          func main() {
+            fmt.Println(add(42, 13))
+          }
+        EOGO
+
+        expect(run_engine(engine_conf)).to be_empty
+      end
```
toddmazierski: Did we confirm that this test fails if the `"(Comment ___)"` filter is removed?

spec/cc/engine/analyzers/go/main_spec.rb
```diff
@@ -0,0 +1,180 @@
...
+            fmt.Println("Add some stuff!")
+          }
+
+          func add(x int, y int) int {
+            return x + y
+            fmt.Println("Add all the stuff!")
+          }
+
+          func main() {
+          	fmt.Println(add(42, 13))
```
toddmazierski: An indentation quibble: there might be a mix of tabs and spaces here and line 66.

spec/cc/engine/analyzers/go/main_spec.rb
```diff
@@ -0,0 +1,180 @@
...
+          import "fmt"
+
+          func main() {
+            fmt.Println("Hello!")
+          }
+        EOGO
+
+        issues = run_engine(engine_conf).strip.split("\0")
+        expect(issues).to be_empty
+      end
```
toddmazierski: I might suggest we modify this test to use two separate source files (ex. `foo.go` and `bar.go`) since this example does not compile:

```
tmp/sandbox319209057/main.go:4:18: fmt redeclared as imported package name
	previous declaration at tmp/sandbox319209057/main.go:3:18
```

FYI: you can easily try out these examples at [The Go Playground](https://play.golang.org/).

larkinscott:
Thanks, @toddmazierski! You were right from the beginning about the comment filtering. The example you mentioned with the PHP duplication engine seemed to do the trick. Would you mind taking a quick look at my changes before I merge?

PR #295, opened 2017-12-08
Removed docstring from Python AST, thereby removing it from analysis

JustinTervala:
Python docstrings were being analyzed in my codebase, which was obscuring important code duplications. Python docstrings don't have a unique element in the AST; rather, triple-quoted strings are interpreted as docstrings by their location and context in the AST. This pull request walks through the nodes in the AST, determines if the node has a docstring in the same way that Python's built-in `ast.get_docstring(node)` does, and if the node has a docstring, it removes the Expr node representing the docstring from the node while preserving the line numbers of the original source file. 

This code was tested in Python 2.7, 3.4, 3.5, and 3.6, against a source file which contained all the possible locations a docstring could be located. If there is somewhere for me to add this test to a suite, I would appreciate it if someone would point in there so I can add a few tests to the test suite.

### DIFF
--- a/lib/cc/engine/analyzers/python/parser.py
+++ b/lib/cc/engine/analyzers/python/parser.py
@@ -1,6 +1,15 @@
-import json, sys, ast
+import json, sys, ast, _ast
 
 PY3 = sys.version_info[0] == 3
+
+try:
+    from _ast import AsyncFunctionDef
+    additional_docstring_types = [AsyncFunctionDef]
+except ImportError:
+    additional_docstring_types = []
+
+possible_docstring_types = [_ast.FunctionDef, _ast.ClassDef, _ast.Module]
+possible_docstring_types.extend(additional_docstring_types)
 
 def string_type():
     return str if PY3 else basestring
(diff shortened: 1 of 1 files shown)


ale7714:
@JustinTervala thank you for using Code Climate and opening this PR. I have 2 comments before continuing forward with the review. 
First, can you open this PR agains the branch `channel/cronopio` instead master? We've been working on this branch for the [new maintainability checks](https://codeclimate.com/blog/10-point-technical-debt-assessment/) that were rollout recently. On the near future `channel/cronopio will be `master` but we didn't want to do that before the release. 
Also, can you add some tests to the PR? 

Again, thank you and let us know if anything.

JustinTervala:
After writing a bunch of tests, I can't seem to replicate the problem I was having using docstrings alone. It appears that similarly-named functions with identical parameter names (e.g. in my case `def pause_workflow(self, task_id, workflow_id)` and `def resume_workflow(self, task_id, workflow_id)`) was causing a lot of false flags in my codebase. If you would like, I can submit tests ensuring that docstrings aren't in the parsed AST, but I don't think it would solve the issue I was experiencing. I'll look at configuring this engine better.

ale7714:
@JustinTervala if you're experiencing an issue that you want to report, you could open an issue on this repo including some snippets of code that can help us reproduce the problem. Also, you can always contact us over at https://codeclimate.com/help.

PR #296, opened 2017-12-08
Fix comment exclusion issue

larkinscott:
This fixes the comment exclusion issue referenced here: https://twitter.com/sdboyer/status/939159860159111169

I was originally trying to exclude comments as you would for a PHP s-exp and not Go, so I wasn't really excluding comments at all.

### DIFF
--- a/lib/cc/engine/analyzers/go/main.rb
+++ b/lib/cc/engine/analyzers/go/main.rb
@@ -18,14 +18,14 @@ module CC
           ].freeze
           POINTS_PER_OVERAGE = 40_000
           REQUEST_PATH = "/go"
-          COMMENT_MATCHER = Sexp::Matcher.parse("(_ (comments ___) ___)")
-
-          def use_sexp_lines?
-            false
-          end
+          COMMENT_MATCHER = Sexp::Matcher.parse("(CommentGroup ___)")
 
           def transform_sexp(sexp)
             delete_comments!(sexp)
+          end
+
+          def use_sexp_lines?
+            false
           end
 
           private
--- a/spec/cc/engine/analyzers/go/main_spec.rb
+++ b/spec/cc/engine/analyzers/go/main_spec.rb
@@ -36,13 +36,13 @@ module CC::Engine::Analyzers
           "path" => "foo.go",
           "lines" => { "begin" => 6, "end" => 6 },
         })
-        expect(json["remediation_points"]).to eq(820_000)
+        expect(json["remediation_points"]).to eq(500_000)
         expect(json["other_locations"]).to eq([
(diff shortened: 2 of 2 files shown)


wfleming:
This looks really close, but glancing at a Go AST locally I think we might need to try and strip out a bit more from the sexp.

Something very small like

```
// @flow
package foo
```

Results in an AST like

```
{
  "comments": [
    {
      "list": [
        {
          "positions": {
            "begin": {
              "column": 1,
              "line": 1
            },
            "end": {
              "column": 10,
              "line": 1
            }
          },
          "text": "// @flow ",
          "type": "Comment"
        }
      ],
      "positions": {
        "begin": {
          "column": 1,
          "line": 1
        },
        "end": {
          "column": 10,
          "line": 1
        }
      },
      "type": "CommentGroup"
    }
  ],
  "decls": null,
  "doc": {
    "list": [
      {
        "positions": {
          "begin": {
            "column": 1,
            "line": 1
          },
          "end": {
            "column": 10,
            "line": 1
          }
        },
        "text": "// @flow ",
        "type": "Comment"
      }
    ],
    "positions": {
      "begin": {
        "column": 1,
        "line": 1
      },
      "end": {
        "column": 10,
        "line": 1
      }
    },
    "type": "CommentGroup"
  },
  "imports": null,
  "name": {
    "name": "foo",
    "positions": {
      "begin": {
        "column": 9,
        "line": 2
      },
      "end": {
        "column": 12,
        "line": 2
      }
    },
    "type": "Ident"
  },
  "type": "File",
  "unresolved": null
}
```

Once we've deleted the `CommentGroup`s, we'll be left with empty `comments` and `doc` attrs that will still contribute to the mass.

So something similar to the PHP one, with an extra layer around the `CommentGroup`, may be necessary?

larkinscott:
After looking at this a little more, I'm not convinced that comment exclusion is the issue here. The original comment exclusion pattern seems to exclude "comments" and "CommentGroup" and also satisfies the tests.

I was able to locate the OSS repo the user complained about: https://github.com/golang/dep. After looking at the two files in question, I'm not really sure how our Duplication engine would flag these two files as similar:

https://github.com/golang/dep/blob/master/gps/deduce.go
https://github.com/golang/dep/blob/master/gps/manager_test.go

Is it possible that, sort of like you mentioned above, we need to exclude a lot more from the AST, beyond just comments?

wfleming:
Scott was right, I was being silly & presuming aspects of the code that are not true,

The comment deleting code doesn't delete the top-level element it matches, it deletes the *first child of the matched thing*. I think that's a confusing choice and binds the comment match pattern and the deletion logic too closely, but the current code seems to be working as intended, so I'm closing this.

PR #297, opened 2017-12-12
Removes duplication definition of `skip?` method

chrishulton:
This fixes a bug where issues were being reported when the thresholds
were not being met, due to a second definition of `skip?` replacing the
correct one.

This also improves the specs to properly cover the behavior.

related: https://github.com/codeclimate/escalations/issues/206

### DIFF
--- a/lib/cc/engine/analyzers/reporter.rb
+++ b/lib/cc/engine/analyzers/reporter.rb
@@ -152,14 +152,6 @@ module CC
           (violation.occurrences + 1) < language_strategy.count_threshold
         end
 
-        def skip?(violation)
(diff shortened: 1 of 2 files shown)


lib/cc/engine/analyzers/reporter.rb
```diff
@@ -152,14 +152,6 @@ def insufficient_occurrence?(violation)
           (violation.occurrences + 1) < language_strategy.count_threshold
         end
 
-        def skip?(violation)
```
chrishulton: i'm assuming this was some kind of bad merge.

PR #299, opened 2017-12-13
Filter out "name" and "unresolved" nodes

larkinscott:
This is in response to issue https://github.com/codeclimate/app/issues/6609. 

@toddmazierski and I paired on this Go duplication bug yesterday afternoon and came up with this fix. It's patchwork and we agree that the real fix should be at the parser level, since it's essentially returning a lot of information we don't need (see: https://github.com/codeclimate/app/issues/6609). 

My inclination would be to merge this in now since the bug is currently live in production, but ultimately fix the parser.

Thoughts @wfleming @katwchen ?

### DIFF
--- a/lib/cc/engine/analyzers/go/main.rb
+++ b/lib/cc/engine/analyzers/go/main.rb
@@ -18,14 +18,14 @@ module CC
           ].freeze
           POINTS_PER_OVERAGE = 40_000
           REQUEST_PATH = "/go"
-          COMMENT_MATCHER = Sexp::Matcher.parse("(_ (comments ___) ___)")
-
-          def use_sexp_lines?
-            false
-          end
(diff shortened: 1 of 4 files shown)


larkinscott:
>Is what was happening here that these nodes don't have real line information, so it's the old issue of when they match it looks like the whole file even though it isn't, since missing line information is (incorrectly) filled in as the whole file by the parser client? 

I'm pretty sure that was the issue here. 

And yeah, that makes sense! I'm happy to take a look at the Go parser and see if I can get a fix working today.

larkinscott:
Updated this PR to reflect only filtering at "name" at the s-exp level. Made a tweak to the parser which excludes "unresolved" from the AST: https://github.com/codeclimate/codeclimate-parser/pull/138.

PR #310, opened 2018-04-19
Fix count_threshold example configuration

kankje:
The custom `count_threshold` example for all languages in the README doesn't work. `count_threshold` should be set inside the `config` key to work properly.


wfleming:
That was a silly mistake, thanks for the correction @kankje

PR #323, opened 2018-06-14
Increase Kotlin duplication mass threshold

larkinscott:
- After looking at some of the duplication issues this was surfacing locally and running some metrics, I think it would be good to bump this from `40` to `60`

### DIFF
--- a/lib/cc/engine/analyzers/kotlin/main.rb
+++ b/lib/cc/engine/analyzers/kotlin/main.rb
@@ -12,7 +12,7 @@ module CC
         class Main < CC::Engine::Analyzers::Base
           LANGUAGE = "kotlin".freeze
           PATTERNS = ["**/*.kt"].freeze
-          DEFAULT_MASS_THRESHOLD = 40
+          DEFAULT_MASS_THRESHOLD = 60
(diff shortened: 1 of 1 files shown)


larkinscott:
@chrishulton I was leaning towards 60 after looking at a range of the issues that were being surfaced, so I decided to try to get some metrics. It's not perfect, but I took a handful of Kotlin repos, dropped them into a directory, set the `mass_threshold` to 20 and collected all the resulting mass_thresholds using this: https://github.com/codeclimate/qm_qa/pull/16/files

The result was:
```
{'metrics': [{'language': 'Kotlin', '90th percentile': 84.0, '95th percentile': 126.0, 'Mean': 55.250245076196414, 'Average': 55.250245076196414, 'Median': 39.0, 'Standard deviatation': 80.4210381927305, 'Max value': 2777, 'Min value': 20}]}
```

I think the mean seems right here. Anything higher would eliminate some actual duplication issues.

chrishulton:
Cool, I'm not really sure what kind of math makes sense here.  If we had good issue status data (mark as confirmed => mark as invalid) this could be a good candidate for machine learning.  Maybe something we can try out later ..........

PR #324, opened 2018-06-21
Add support to scala

filipesperandio:
RE: https://github.com/codeclimate/app/issues/7173

### DIFF
--- a/lib/cc/engine/analyzers/scala/main.rb
+++ b/lib/cc/engine/analyzers/scala/main.rb
@@ -13,7 +13,9 @@ module CC
           LANGUAGE = "scala".freeze
           PATTERNS = ["**/*.sc", "**/*.scala"].freeze
           DEFAULT_MASS_THRESHOLD = 60
-          DEFAULT_FILTERS = [].freeze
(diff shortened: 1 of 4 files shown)


lib/cc/engine/analyzers/scala/main.rb
```diff
@@ -0,0 +1,37 @@
...
+
+module CC
+  module Engine
+    module Analyzers
+      module Scala
+        class Main < CC::Engine::Analyzers::Base
+          LANGUAGE = "scala".freeze
+          PATTERNS = ["**/*.sc", "**/*.scala"].freeze
+          DEFAULT_MASS_THRESHOLD = 60
+          DEFAULT_FILTERS = [].freeze
```
wfleming: Do you think we need to ignore imports by default? That's a common source of false positives in many languages.

filipesperandio: I think I haven't hit issues with the current `DEFAULT_MASS_THRESHOLD`. Improving the spec so I can adjust this! 👍

PR #329, opened 2018-07-20
Remove docker login step from Circle yml

larkinscott:
This removes the docker login step from our Circle yml. The idea is
that this is pulling the parser image, which was private when we added
this step. However, the parser image is now public.

This is also blocking OSS contributions from users who have forked
duplication, since Circle fails at this step on forks.

Co-authored-by: Max Jacobson <max@hardscrabble.net>


maxjacobson:
That theory sounds right to me, and the fact that this is green seems like pretty good evidence that the line isn't necessary. Seems like we should merge this, then encourage OSS contributors to rebase on top of master

PR #333, opened 2018-09-14
Add .py3 extension to python file patterns

dblandin:
Our Python parser already supports both Python 2 and Python 3 source code. While most Python 3 files share the same `.py` file extension used for Python 2 files, some projects use the `.py3` file extension.


wfleming:
It should be noted this may have some unexpected results: this engine supports python 3 *with explicit configuration*: see when support was introduced in https://github.com/codeclimate/codeclimate-duplication/pull/135. As-is this will try to run the python2 parser against python 3 files by default, which may lead to unexpected results.

PR #337, opened 2018-10-01
README: update installation

wfleming:
There is no `engines:enable` command anymore, and duplication is on by default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ simple. We have a [docs page][what-is-duplication] explaining the algorithm.
 ## Installation
 
 1. Install the [Code Climate CLI][cli], if you haven't already.
-3. You're ready to analyze! `cd` into your project's folder and run `codeclimate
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -14,10 +14,9 @@ simple. We have a [docs page][what-is-duplication] explaining the algorithm.
 ## Installation
 
 1. Install the [Code Climate CLI][cli], if you haven't already.
-2. Run `codeclimate engines:enable duplication`. This command installs the
-   engine and enables it in your `.codeclimate.yml` file.
 3. You're ready to analyze! `cd` into your project's folder and run `codeclimate
```
maxjacobson: Maybe we should change all of these numbers to `1. ` to avoid this numeric gap. We could also consider rewriting this into a paragraph rather than a list.

wfleming: Good catch. I went with option 1 for simplicity for now.

PR #353, opened 2020-01-08
Bump `ruby_parser` to `3.14.1`

larkinscott:
This bumps `ruby_parser` to resolve an escalation with regards to how
the duplication engine handled some more modern Ruby
syntax: https://github.com/codeclimate-support/quality/issues/12.

This bump also exposes a bug that was introduced into how the parser
handles heredocs, so marking a test that has become less-than-helpful
due to that as pending.


spec/cc/engine/analyzers/ruby/main_spec.rb
```diff
@@ -140,6 +141,19 @@ def self.from_remediation_amount(amount)
         expect(issues.length).to eq(3)
       end
 
+      it "parses modern (>2.6) Ruby syntax" do
```
larkinscott: This fails to parse before the parser version bump.

spec/cc/engine/analyzers/sexp_lines_spec.rb
```diff
@@ -18,9 +18,9 @@ module CC::Engine::Analyzers
 
         expect(locations.count).to eq 2
         expect(locations[0].begin_line).to eq(3)
-        expect(locations[0].end_line).to eq(7)
+        expect(locations[0].end_line).to eq(4)
```
larkinscott: I _think_ this is a product of some tweaks to the way the parser handles these situations.

PR #357, opened 2021-01-29
Bump cc-parser to b850

fede-moya:


fede-moya:
@filipesperandio same issue here of shipbot not receiving notification.

PR #364, opened 2021-06-16
Python: don't crash on complex-number literals

jakelishman:
This fixes #363: crashing on complex-number literals in Python 2 and 3.  I modified the JSON serialiser to string-ify complex numbers, rather than passing them directly as numeric values.  That's perhaps not 100% ideal, but it's the same as the behaviour of Ruby's `to_json` method.  I've added a test that the behaviour works, as well.

Running `codeclimate analyze` on the same test file in issue #363 with the same settings (`mass_threshold: 1`) now produces the expected output in both Python 2 and Python 3:
```
Starting analysis
Running duplication: Done!

== complex.py (3 issues) ==
1-2: Similar blocks of code found in 3 locations. Consider refactoring. [duplication]
4-5: Similar blocks of code found in 3 locations. Consider refactoring. [duplication]
7-8: Similar blocks of code found in 3 locations. Consider refactoring. [duplication]

Analysis complete! Found 3 issues.
```

PR #366, opened 2021-08-10
Do not run relase-images on forked prs

fede-moya:
We are receiving failed builds from prs open by the community due to running the `relase-images` ci job. This work removes this job for those cases.

See https://github.com/codeclimate/codeclimate-duplication/pull/364


fede-moya:
This theory was wrong the CI was building due to a missing dependency on the `codelcimate-parser` base image.

filipesperandio (commented):
Skipping CI for those, how we are going to validate them instead?

fede-moya:
> Skipping CI for those, how we are going to validate them instead?
@filipesperandio it's not the entire ci, just the job for releasing images. Do you think releasing images is a necessary step for validating prs ?

filipesperandio:
🤦 I was tired...

filipesperandio (approved):
I am OK shipping this anyways, but seems like we only actually release images for master and channel-* branches here: https://github.com/codeclimate/codeclimate-duplication/blob/master/bin/deploy#L10-L13

fede-moya:
> I am OK shipping this anyways, but seems like we only actually release images for master and channel-* branches here: https://github.com/codeclimate/codeclimate-duplication/blob/master/bin/deploy#L10-L13

mmmm, interesting because it is getting triggered on https://github.com/codeclimate/codeclimate-duplication/pull/366

filipesperandio (approved):
Yeah, that build seems to break on `docker login`, even before the `bin/deploy` call, but I agree this is even better as we skip the step entirely, have CIRCLE env vars checked on Circle related files and make it more visible.

👍

fede-moya:
@filipesperandio Exactly it was breaking before the `bin/deploy`, could you merge it ? 🙏🏼

